PR #12, opened 2015-06-17
public boolean connectToDriverAtWithCapabilities(String url, String capabilitiesInJson)

sglebs:
With JSON I can type each capability (not just everything is a string)


fhoeben:
Have you tested this (and the old way using a Map)? I believe Slim will not be able to handle the fact that there are now two methods with the same name and the same number of parameters (but different types).

PR #23, opened 2015-10-01
Multiple changes: 1. enable default value and 2. enable comparing returning cell value with text containing symbol and 3. Ftp file utilities

ilseh:
1. Enable default value: to facilitate in fixture to have an optional column, using a default value if none is specified in table.
2. Enable comparing return value of a column with text value that contains symbols. Example of use is given in wikipage: HsacExamples.FitTests.ArraysAndSymbolsComparison#Compare result with string containing a symbol


src/main/java/nl/hsac/fitnesse/fixture/fit/MapColumnFixture.java
```diff
@@ -376,6 +395,26 @@ public void doCell(Fixture aFixture, Parse aCell) {
...
+    /**
+     * Functionality to compare returning cell value with a dynamic (with symbols) text.
+     * See also HsacExamples.FitTests.ArraysAndSymbolsComparison#Compare result with string containing a symbol
+     * @param originalCellText, ie randomText$[symbol1]MoreText where symbol1=ValueOfSymbol1
+     * @return originalCellText with the symbols resolved to values, ie: randomTextValueOfSymbol1MoreText
+     *      or null if the originalCellText doesn't contains text with symbols
+     * @throws NoSuchSymbolException if the used symbol(s) don't exist
+     */
+    private String resolveStringWithSymbols(String originalCellText) throws NoSuchSymbolException {
+        Pattern p = Pattern.compile(REGEX_STRING_WITH_SYMBOL);
```
fhoeben: Why don't you store the pattern in a final static (instead of only the String and compiling each time)?

fhoeben:
I don't see a sample of the default value feature being used. Is that correct?

src/main/java/nl/hsac/fitnesse/fixture/fit/MapColumnFixture.java
ilseh: Thanks for the suggestion, i will change it.

ilseh:
There is indeed no example of the default value. We use it the execute method of a fixture, like this:
```
public class ReadEmailFixture extends MapColumnFixture {
    public static final String DEFAULT_SERVER_PROTOCOL = "imaps";
    private static final String PROTOCOL_KEY = "protocol";
...
   @Override
    public void execute() {
        ..
        String mailServerProtocol = (String) get(PROTOCOL_KEY, DEFAULT_SERVER_PROTOCOL);
..
```

fhoeben:
Regarding the default value: why was a separate method needed instead of just overriding MapColumnFixture's?
```
    protected void setDefaults(Map<String, Object> values) {
    }
```
That method is especially there to support setting default values for sub classes...

ilseh:
Removed the 
``` public Object get(String columnName, Object defaultValue) ```
as suggested because it was totally unnecessary. (References use setDefaults now instead.)

ilseh:
Added ReadFileFromFtpServerFixture that uses the FtpFileUtil and added an example on how to use it.

PR #25, opened 2015-10-05
Possibility to set profile when firefox is used. Used to be able to download files in test on buildserver without a dialog box interfering.

ilseh:


fhoeben:
Setting a profile could be a very good feature. 

Do you only use it for a download directory (as I see in the code), or do you need more? What other features does this give you?

Did you see there are special methods in BrowserTest to do downloads (that work cross browser and in a Selenium grid setup)?

ilseh:
We use the profile currently only to make firefox download a csv file after pressing a button- without first showing a dialog window.
I guess we could use the "download content from" instead, we weren't aware of the feature.

But i think the firefox profile could be used for other things, it's written in a generic way.

A second look on the 
```
    public boolean startDriver(String driverClassName, final Map<String, String> profile) throws Exception {
       ..
        SeleniumHelper.DriverFactory driverFactory = new SeleniumHelper.DriverFactory() {
            @Override
            public void createDriver() {
             ..
                        FirefoxProfile fxProfile = getFireFoxProfile(profile);

                        final String downloadDir = new LocalSeleniumDriverFactoryFactory().getProperty(SELENIUM_DOWNLOAD_DIRECTORY);
                        if (StringUtils.isNotBlank(downloadDir)) {
                          fxProfile.setPreference("browser.download.dir", downloadDir);
                        }
```
made me realize the browser.download.dir shouldn't be set in the code because it should be set in the profile wiki and this suggests it's download specific so i removed that last part.

src/main/java/nl/hsac/fitnesse/junit/selenium/SeleniumDriverFactoryFactoryBase.java
```diff
@@ -23,7 +23,7 @@ protected boolean isPropertySet(String propertyName) {
         return !StringUtils.isEmpty(value);
     }
 
-    protected String getProperty(String propertyName) {
+    public String getProperty(String propertyName) {
```
fhoeben: Is this still needed after removing the download stuff?

src/main/java/nl/hsac/fitnesse/fixture/slim/web/SeleniumDriverSetup.java
```diff
@@ -94,27 +112,31 @@ public boolean startDriverFor(String browser) throws Exception {
         boolean result = false;
         String browserName = browser.toLowerCase();
         if ("firefox".equals(browserName)) {
-            result = startDriver(FirefoxDriver.class.getName());
+            result = startDriver(FirefoxDriver.class.getName(), profile);
         } else if ("safari".equals(browserName)) {
-            result = startDriver(SafariDriver.class.getName());
+            result = startDriver(SafariDriver.class.getName(), profile);
```
fhoeben: Why do we pass a profile for all drivers, if only the firefox will pick it up?

src/main/java/nl/hsac/fitnesse/fixture/slim/web/SeleniumDriverSetup.java
```diff
@@ -198,6 +220,32 @@ public boolean connectToDriverAtWithCapabilities(String url, Map<String, Object>
...
+    }
+
+    /**
+     * Set firefox profile. For example to make sure text/csv file is downloaded without asking (convenient if run on buildserver), do:
+     * |script           |selenium driver setup                                                                                               |
+     * |start driver for |firefox              |with profile|!{browser.download.folderList:2,browser.helperApps.neverAsk.saveToDisk:text/csv}||
+     * @param profile setting from subtable
+     * @return firefox profile with specified settings
+     */
+    private FirefoxProfile getFireFoxProfile(Map<String, String> profile) {
```
fhoeben: Shouldn't this method be called getFirefoxProfile (no capital F in fox)?

src/main/java/nl/hsac/fitnesse/fixture/slim/web/SeleniumDriverSetup.java
```diff
@@ -198,6 +220,32 @@ public boolean connectToDriverAtWithCapabilities(String url, Map<String, Object>
         return createAndSetRemoteDriver(url, desiredCapabilities);
     }
 
+    public boolean connectToFireFoxDriverAtWithProfile(String url, Map<String, String> profile)
```
fhoeben: Shouldn't this be called connectToFirefoxDriverAtWithProfile (no capital F in fox)

PR #26, opened 2015-10-05
Add support for HTTP verb PUT

### LINKED ISSUE #20: Do you support PUT
If not would you have any plans to add it?

sbarne3:
fixes #20

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
@@ -15,7 +15,7 @@ import java.util.regex.Pattern;
  * Fixture to make HTTP requests using Slim scripts and/or scenarios.
  */
 public class HttpTest extends SlimFixtureWithMap {
-    /** Default content type for posts. */
+    /** Default content type for posts and puts. */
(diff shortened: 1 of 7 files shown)


src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
```diff
@@ -257,6 +314,14 @@ protected String urlEncode(String str) {
     protected boolean postProcessResponse() {
         return responseIsValid();
     }
+    
+    /**
+     * Performs any put processing directly after retrieving response.
+     * @return true if all is well, false otherwise.
+     */
+    protected boolean putProcessResponse() {
```
fhoeben: I did not expect this method. Would calling postProcessResponse() not be ok also. The 'post' in postProcessResponse did not really refer to http's 'POST' but more to the general meaning of 'after'...

sbarne3: Ah... that makes sense. Let me change that.

fhoeben:
Looks very good. Could you also add the new commands to the documentation (HsacExamples.SlimTests.HttpTests), and ideally also add an example test page (or add an example to HsacExamples.SlimTests.HttpTests.JsonHttpTest)?

PR #30, opened 2015-10-07
WikiHelper - expose wiki's files section also to fit fixtures

ilseh:
I added a WikiHelper that can also be used by fit fixtures in case usage of wiki's files section is necessary. Eventually we didn't need it so this functionality is not used currently. Some minor refactoring was done for this (also in SlimFixture).

What is your opinion on this?


ilseh:
Originally i needed this to test camel flows that pass around files. I wanted to put the test files in the files section of wiki. The setup of the fitnesse test could then copy the test file from the files section to the starting point of the camel flow.

fhoeben:
I was referring specifically to the extra logic in `Environment.getFitNesseFilesSectionDir()`, why was that needed? I never needed that, either on build server or locally.

fhoeben:
I understand we can expose the basic behavior also to Fit. I believe a FilesSectionsHelper might be a better name for the class, though. This can then also have the feature to determine a subdirectory below the files section, which is now repeated in multiple places (in field initializers)


(FYI: To create, and copy, files (possibly in/from files section) slim tests can now (in the 2.2.0 release) also use the FileFixture.)

ilseh:
I need the ```Environment.getFitNesseFilesSectionDir``` because when i run it (in Eclipse -> Maven Build: clean compile exec:exec) its current path is: {SOURCE_DIR_FITNESSE}\wiki, When i run it with the command to run it on the buildserver (in Eclipse -> Maven Build: clean test-compile failsafe:integration-test) it's current path is {SOURCE_DIR_FITNESSE}. Maybe there is something wrong in my setup.

fhoeben:
A run on the build server should set the path for the files section explicitly (by `Environment.getInstance().setFitNesseRoot(outputDir);` in `HsacFitNesseRunner`). In my usage I want the files section to be in '{SOURCE_DIR_FITNESSE}/target/fitnesse-results' so all files created during my tests are part of the test report generated by the build server. So I can see the screenshots generated by `storyboard`s and files downloaded by `BrowserTest`, I don't want that to change!

I don't place files in the files section manually, but the files are created (by fixtures) during the test run and then potentially picked up by other fixtures.
Do you have manually placed files in the files section that you want your fixtures to pick up? Maybe the maven resources plugin's `copy-resources` could help with that?

ilseh:
Ok, since we don't really need this,. Fixing the broken screenshots path without breaking other things isn't worth the effort. 
(We indeed would put the files in the files section per hand.)

PR #43, opened 2016-02-09
followRedirect

AukeRoze:
new option in Get to stop followRedirect


fhoeben:
Can you also update the documentation (in HsacExamples.SlimTests.HttpTests) to include this new option?

AukeRozema:
Yes thanks,

I was looking for that ☺

I updated the documenation.

fhoeben:
Did you push that update to the documentation? I don't see it in this pull request...

AukeRoze:
I pushed this in the location you asked for it.



Date: Thu, 11 Feb 2016 06:43:43 -0800

fhoeben:
Hi Auke,

Great to see you added an example test. I was also hoping you would update the table in HsacExamples.SlimTests.HttpTests that lists the command supported by HttpTest.

Furthermore I think it would be nicer if the disabling of redirects was a setting/method that would be applied on HttpTest as a whole and not an overload (aka with extra parameter) on getFrom. I mean that you could do:

```
|set follow redirects|false|
|get from|${URL}|
```

Do you have a special reason to do it as an overload?

wiki/FitNesseRoot/HsacExamples/SlimTests/HttpTests/HttpGetWithRedirectTest/content.txt
```diff
@@ -0,0 +1,9 @@
+!2 Http GET without follow redirect
+
+Performs HTTP GET adding values as parameters in the query string.
+
+|script         |xml http test                                                       |
+|get from       |${URL}         | with redirect       | false        |
```
fhoeben: Do you know a public site that actually sends a redirect we could use here.
I prefer all the examples to be green...

src/main/java/nl/hsac/fitnesse/fixture/util/HttpClient.java
```diff
@@ -1,35 +1,73 @@
...
+                        }
+                        return isRedirect;
+                    }
+
+
+                })
+                .build();
+    }
+
+    public URI createLocationURI(final String location) throws ProtocolException {
```
fhoeben: Why did you create this method? I see no usage of it.

src/main/java/nl/hsac/fitnesse/fixture/util/HttpClient.java
```diff
@@ -1,35 +1,73 @@
...
-                .setUserAgent(HttpClient.class.getName()).build();
+
+        HTTP_CLIENT = HttpClients.custom()
+                .useSystemProperties()
+                .disableContentCompression()
+                .setUserAgent(HttpClient.class.getName())
+                .setRedirectStrategy(new DefaultRedirectStrategy() {
+                    @Override public boolean isRedirected(HttpRequest request, org.apache.http.HttpResponse response, HttpContext context) throws ProtocolException {
+                        boolean isRedirect = super.isRedirected(request, response, context);
+                        if(isRedirect && !followRedirect) {
```
fhoeben: To me this if statement is not very clear. It probably is the name of the superclass' method, or the local variable.

Can you may make a bit clearer (by naming or comments) what the various cases are/mean?

Furthermore I would prefer it if the method only has a single return statement. I find that makes it a bit easier to understand the flow.

Issue #47, opened 2016-03-21
Request: extend fixture to support appium?


tcnh:
Not an issue, more like a feature request.

Would it be possible to extend the driver setup and helpers to make use of AndroidDriver (appium) and support MobileElements (subclass of WebElement, but with bonus-operations), so that we could create a native app test fixture that supports swiping, multi-touch and mobile element locating?

I tried extending it, but alas my experience is insufficient. (I get an AndroidDriver, but then fail, because the rest of the fixtures are expecting WebElements and not MobileElements). Using only standard webDriver commands, automating native apps already works quite nicely using your fixtures.

fhoeben:
What would be be needed to make use of AndroidDriver? I would expect it already to be possible, but maybe a bit cumbersome.

Can you maybe share your attempt (via a fork) and explain where you run into problems?

tcnh:
Got it running, still needs a bit of polishing. I'll create a demo page + a small native app fixture to show the extra functionality and share in a fork.

tcnh:
New approach in https://github.com/tcnh/hsac-fitnesse-fixtures/tree/androidFixture
Not ready to be a PR yet, but I think this is a better way to integrate Appium than the initial approach. We now have two new fixtures (one for iOS and one for android), both implementing the same interface to allow for cross platform scripting.

Element locating uses a very basic level of heuristic to determine what element to use (for Android. iOS fixture uses the first one that matches the lookup). Locating requires minimal technical knowledge of the application, as it tries to find elements by visible text, decriptions, hints or id-partials. Will try and find a way of adding acceptance tests that work with sauce in the next days and also create documentation.

Would appreciate any feedback or suggestions

fhoeben:
Looks cool feature-wise. But it is quite a bit of code. I have to find some time to look through it properly!

fhoeben:
I've started work (based on your ideas) in a new project. Any help/suggestions are welcome: https://github.com/fhoeben/hsac-fitnesse-appium

PR #48, opened 2016-04-11
Extended fixture set with appium android native app fixture

tcnh:
Added androidAppFixture
Extended SeleniumDriverSetup with possibility to ask for an Appium AndroidDriver 
Some extensions to SeleniumHelper (quick and -sometimes very- dirty)

Some documentation and examples in FitNesse wiki

#47

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/mobile/AndroidAppTest.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/mobile/AndroidAppTest.java
@@ -1,0 +1,221 @@
+package nl.hsac.fitnesse.fixture.slim.mobile;
+
+import io.appium.java_client.*;
+
+import io.appium.java_client.android.AndroidDriver;
+import nl.hsac.fitnesse.fixture.slim.SlimFixtureException;
+import org.openqa.selenium.*;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.WebDriverWait;
+
+import io.appium.java_client.remote.MobileCapabilityType;
+import io.appium.java_client.remote.MobilePlatform;
+
+import nl.hsac.fitnesse.fixture.slim.web.BrowserTest;
+import nl.hsac.fitnesse.fixture.slim.web.annotation.TimeoutPolicy;
+import nl.hsac.fitnesse.fixture.slim.web.annotation.WaitUntil;
+
+import java.lang.reflect.Method;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Android App fixture for use with an android driver. Aimed at effortlessly testing native apps
+ */
+
+public class AndroidAppTest extends BrowserTest {
+    private int SWIPESPEED = 1000;
+    private int WAITAFTERSWIPE = 0;
+
+    public AndroidAppTest() {
+        super("X");
+    }
+
+    @Override
+    protected void beforeInvoke(Method method, Object[] arguments) {
+        //intentionally left empty..
+    }
+
+    public void setImplicitWaitInSeconds(long stepDelay){
+        getSeleniumHelper().androidDriver().manage().timeouts().implicitlyWait(stepDelay, TimeUnit.SECONDS);
+    }
+
+    public void setSwipeSpeedInMilliseconds(int swipeSpeed){
+        this.SWIPESPEED = swipeSpeed;
+    }
+
+    public void setWaitAfterSwipe(int waitAfterSwipe){
+        this.WAITAFTERSWIPE = waitAfterSwipe;
+    }
+
+    @WaitUntil(TimeoutPolicy.RETURN_FALSE)
+    public boolean tap(String locator){
+
+        boolean result = false;
+        MobileElement element = getMobileElement(locator);
+
+        if(element != null){
+            TouchAction tap = new TouchAction(getSeleniumHelper().androidDriver()).tap(element);
+            tap.perform();
+
+            result = true;
+        }
+
+        return result;
+    }
+
+    @WaitUntil(TimeoutPolicy.RETURN_FALSE)
+    public boolean enterAs(String value, String locator) {
+        boolean result = false;
+        MobileElement element = getMobileElement(locator);
+
+        if(element != null){
+            element.sendKeys(value);
+            getSeleniumHelper().androidDriver().hideKeyboard();
+            result = true;
+        }
+
+        return result;
+    }
+
+    @WaitUntil(TimeoutPolicy.RETURN_NULL)
+    public String valueOf(String locator){
+        MobileElement el = getMobileElement(locator);
+        return el.getText();
+    }
+
+    @WaitUntil(TimeoutPolicy.RETURN_FALSE)
+    public Boolean elementDisplayed(String locator){
+        return(getMobileElement(locator).isDisplayed());
+    }
+
+    public void swipeUp() {
+
+        Dimension size = getSeleniumHelper().androidDriver().manage().window().getSize();
+        int startX = size.width / 2;
+        int startY = (int) (size.height * 0.8);
+        int endY = (int) (size.height * 0.2);
+
+        swipe(startX, startY, startX, endY, SWIPESPEED, WAITAFTERSWIPE);
+    }
+
+    public void swipeDown() {
+
+        Dimension size = getSeleniumHelper().androidDriver().manage().window().getSize();
+        int startX = size.width / 2;
+        int startY = (int) (size.height * 0.2);
+        int endY = (int) (size.height * 0.8);
+
+        swipe(startX, startY, startX, endY, SWIPESPEED, WAITAFTERSWIPE);
+    }
+
+    public void swipeLeft() {
+
+        Dimension size = getSeleniumHelper().androidDriver().manage().window().getSize();
+        int startX = (int) (size.width * 0.8);
+        int endX = (int) (size.width * 0.2);
+        int startY = size.height / 2;
+
+        swipe(startX, startY, endX, startY, SWIPESPEED, WAITAFTERSWIPE);
+    }
+
+    public void swipeRight() {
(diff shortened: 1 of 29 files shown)


fhoeben:
Looks quite extensive. I will have to find some time to look at it properly. Thanks a lot for contributing it.

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
```diff
@@ -37,17 +40,10 @@
                     "  rect.right <= (window.innerWidth || document.documentElement.clientWidth));\n" +
             "} else { return null; }";
 
-    private static final String TOP_ELEMENT_AT =
```
fhoeben: Did you intentionally remove the 'is on top' functionality? Or did you just base your code on an earlier version that did not have it yet, and did something go wrong in a merge?

If it is intentional please describe why. We will have to find a way to make this behavior optional, because it is very desirable in some standard browser based testing.

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
```diff
@@ -37,17 +40,10 @@
...
-                    "  var rect = arguments[0].getBoundingClientRect();\n" +
-                    "  var x = (rect.left + rect.right)/2;\n" +
-                    "  var y = (rect.top + rect.bottom)/2;\n" +
-                    "  return document.elementFromPoint(x,y);\n" +
-                    "} else { return null; }";
-
     private final List<WebElement> currentIFramePath = new ArrayList<WebElement>(4);
     private DriverFactory factory;
     private WebDriver webDriver;
+    private AndroidDriver<MobileElement> androidDriver;
```
fhoeben: Would it be at all possible to store the androidDriver in the webDriver field? I believe it implements the WebDriver interface. 

It looks like this class could remain pretty much the same as it was then. It would then be the caller's responsibility to cast the driver() to AndroidDriver. But your code requires the caller to call different methods so that seems workable...

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
tcnh: This was unintentional. May have been based off earlier version.

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
tcnh: This is where I got stuck at first. I was convinced I'd be able to do that, but didn't get it to work. Current solution works, but I'm convinced it can be done with less code and less overhead.

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -175,6 +175,11 @@ public BrowserTest(int secondsBeforeTimeout) {
         ensureActiveTabIsNotClosed();
     }
 
+    public BrowserTest(String appium){
+
```
fhoeben: Setting seconds before timeout, and ensuring the active tab is not closed does not work for Appium?
I believe that it would probably be better to ensure those methods don't do anything for an Android App test, instead of having the extra constructor

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -1357,7 +1362,7 @@ public String takeScreenshot(String basename) {
         return screenshotFile;
     }
 
-    private String getScreenshotLink(String screenshotFile) {
+    public String getScreenshotLink(String screenshotFile) {
```
fhoeben: Making this protected would suffice, right? This is not something to be called from the wiki.

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -1374,7 +1379,7 @@ private String getScreenshotLink(String screenshotFile) {
         return screenshotFile;
     }
 
-    private String createScreenshot(String basename) {
+    public String createScreenshot(String basename) {
```
fhoeben: Making this protected would suffice, right? This is not something to be called from the wiki.

fhoeben:
Is there any way we can make the android examples actually work? I have some ideas to simplify/cleanup some of the code. But how do I verify it works? I don't believe I can run the actual tests as-is. 

There is a section of the wiki dedicated to automated acceptance tests of the fixtures. These tests can use SauceLabs, which I believe [also supports testing with Appium](https://wiki.saucelabs.com/display/DOCS/Requirements+for+Android+and+iOS+Mobile+Application+Testing#RequirementsforAndroidandiOSMobileApplicationTesting-AndroidRequirements). Can we setup a (small) testset to verify whether the AndroidAppTest does what is intended?

src/main/java/nl/hsac/fitnesse/fixture/slim/mobile/AndroidAppTest.java
```diff
@@ -0,0 +1,221 @@
...
+            lookFor = strategy[0];
+            content = strategy[1];
+        }
+
+        return driver.findElementByAndroidUIAutomator("new UiSelector()." + lookFor + "(\"" + content + "\")");
+
+    }
+
+    @Override
+    public String takeScreenshot(String basename) {
```
fhoeben: Why is this override here? It does exactly the same as the method it overrides.

src/main/java/nl/hsac/fitnesse/fixture/slim/mobile/AndroidAppTest.java
```diff
@@ -0,0 +1,221 @@
...
+    private int SWIPESPEED = 1000;
+    private int WAITAFTERSWIPE = 0;
+
+    public AndroidAppTest() {
+        super("X");
+    }
+
+    @Override
+    protected void beforeInvoke(Method method, Object[] arguments) {
+        //intentionally left empty..
```
fhoeben: Why is it left empty. Just to ensure no waiting for angular?
It would be easier/clearer to, in the constructor just call `setImplicitWaitForAngularTo(false)`

fhoeben:
I've done some code cleanup, that I don't expect to change anything functional, in the two BrowserTest and the subclass (in the [tcnh branch in my repo](https://github.com/fhoeben/hsac-fitnesse-fixtures/tree/tcnh)).
Can you pull this and verify whether this indeed still works? I will than look at the SeleniumHelper stuff.

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
tcnh: Agree, but somehow the timeout setting was called even though I had it overridden in the AndroidAppTest class.

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
tcnh: Protected would suffice. iirc I dit this because I couldn't figure out a good way to detect what driver we needed for the screenshot (see the very ugly stacktrace-check in seleniumHelper).

src/main/java/nl/hsac/fitnesse/fixture/slim/mobile/AndroidAppTest.java
tcnh: As above. I needed the helperclass to know what class was asking for the screenshot. That definitely needs work.

src/main/java/nl/hsac/fitnesse/fixture/slim/mobile/AndroidAppTest.java
tcnh: I guess that would be clearer, yes. The timeout/wait settings throw a lot of errors in appium

tcnh:
Screenshotting no longer works, because the override was deleted. (needed that to determine if takeScreenshot in seleniumhelper had to screenshot androiddriver or webdriver) There should be a better way for that.

I tried creating a saucelabs open sauce account, but my password is not accepted when I want to display my api key to use with an appium test.. so no cigar just yet.

If you have an android device or emulator and the android sdk installed, you can install appium and run the NegenTwee test (see instructions and apk location in test case in wiki) or thuisbezorgd test

fhoeben:
I pushed a commit attempting to fix the screenshot hack with less code (created an override just calling the superclass' method). Does that do the trick?

I'll see what I can manage with my sauce labs account...

wiki/FitNesseRoot/HsacExamples/SlimTests/AndroidAppTests/content.txt
```diff
@@ -0,0 +1,46 @@
...
+More info: [[Google documentation][http://developer.android.com/reference/android/support/test/uiautomator/UiSelector.html]]
+
+Example test cases:
+>ThuisBezorgdDemo (test Thuisbezorgd app installed on device)
+>NegenTweeAppDemo (install 9292 apk on device and run test)
+
+!2 Language
+!3 !-AndroidAppTest-!
+The main commands/keywords (i.e. public methods) offered by !-nl.hsac.fitnesse.fixture.slim.mobile.AndroidAppTest-! are listed below.
+The heuristic to find places tries a number of translations of end user terms to Selenium 'find' commands (e.g. is a full or partial label given, a link's or button's text, a placeholder). The first find that results in an element on the page is used by !-BrowserTest-!.
```
fhoeben: Does the fixture have a heuristic (yet)? From the code and the text above I believe it either uses an exact match (default) or another locator. 

A heuristic, to me, would be something like first trying an exact match on text, then an exact match on accessibility identifier, and then partial matches on these.

Issue #59, opened 2016-05-12
Req: Is Visible on website


teunisnl:
Simply: ability to check if text/element is visually present anywhere on current page (regardless if its currently on the screen)

Long:
The current | is visible | has its perks if you wanna make sure if an element is on screen.
but its more often I would prefer the option | is visible | would have a build in |scroll to|
for compatibility a |is visible on page| would be great.
and yeah i know i could solve this by addind a scroll to, so it would be more perk then a must have

fhoeben:
I understand what you are looking for, its not very hard to add. Do you have the time to create a pull request? 

As workaround (besides scrolling yourself):
- Would 'is enabled' help you at all? I don't believe that requires the element to be scrolled into view?

teunisnl:
Is enable could invoke false errors, so it not a route i would prefer.
A pull request, time I have, but would be first so it will take some time to figure out how to do it.

teunisnl:
#60

PR #60, opened 2016-05-23
Done: Is Visible on Page

teunisnl:
#59


fhoeben:
Looks great. Like the fact the acceptance test has been added!

* Can you also add the new method to the table in HsacExamples.SlimTests.BrowserTests that lists all methods?
* And (little bit of Java styling thing) could you change 'getisvisible()' to:
```
    protected boolean isVisibleImpl(String place, boolean checkOnScreen) {
        boolean result = false;
        WebElement element = getElementToCheckVisibility(place);
        if (element != null && element.isDisplayed()) {
            if (checkOnScreen) {
                result = isElementOnScreen(element);
            } else {
                result = true;
            }
        }
        return result;
    }
```

teunisnl:
it should be updated now, that best description I could come up with for is visible on page

PR #61, opened 2016-05-28
Added support for files as mock responses, added method to update val…

tcnh:
- Added support for files (charset assumes utf-8) as mock responses.
- Added method to update values in json strings.

 Plus acceptance tests.

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/JsonHttpTest.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/JsonHttpTest.java
@@ -74,7 +74,8 @@ public class JsonHttpTest extends HttpTest {
     public void setJsonPathTo(String path, String value){
         String jsonStr = getResponse().getResponse();
         String jsonPath = getPathExpr(path);
-        getResponse().setResponse(pathHelper.updateJsonPathWithValue(jsonStr, jsonPath, value));
+        String newResponse = pathHelper.updateJsonPathWithValue(jsonStr, jsonPath, value);
+        getResponse().setResponse(newResponse);
     }
 
     protected String getPathExpr(String path) {
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/JsonHttpTest.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/JsonHttpTest.java
@@ -66,6 +66,20 @@ public class JsonHttpTest extends HttpTest {
         return result;
     }
 
+    /**
+     * Update a value in a supplied json string by supplied jsonPath
+     * @param path the jsonPath to locate the key whose value needs changing
+     * @param value the new value to set
+     * @param jsonStr the json to update (this may be set to 'response' to use jsonHttpTest's last reponse body
+     */
+    public String setJsonPathToIn(String path, String value, String jsonStr){
+        if(jsonStr.equalsIgnoreCase("response")){
+            jsonStr = getResponse().getResponse();
+        }
+        String jsonPath = getPathExpr(path);
+        return pathHelper.updateJsonPathWithValue(jsonStr, jsonPath, value);
+    }
+
     protected String getPathExpr(String path) {
         String jsonPath = path;
         if (!path.startsWith("$")) {
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/JsonHttpTest.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/JsonHttpTest.java
@@ -67,17 +67,14 @@ public class JsonHttpTest extends HttpTest {
     }
 
     /**
-     * Update a value in a supplied json string by supplied jsonPath
+     * Update a value in a the response by supplied jsonPath
      * @param path the jsonPath to locate the key whose value needs changing
      * @param value the new value to set
-     * @param jsonStr the json to update (this may be set to 'response' to use jsonHttpTest's last reponse body
      */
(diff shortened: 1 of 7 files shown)


wiki/FitNesseRoot/HsacExamples/SlimTests/HttpTests/content.txt
```diff
@@ -44,10 +44,11 @@ The main commands/keywords (i.e. public methods) offered by !-nl.hsac.fitnesse.f
...
-|all json path matches <expression>|Evaluates the expression against the last response received and returns all results. Intended to be used with 'show' command.|
+|post values as json to <url>                   |Sends a POST containing all values (as JSON properties) to the url.                                                                            |
+|json path <expression>                         |Evaluates the (definite) expression against the last response received.                                                                        |
+|json path count <expression>                   |Evaluates the (indefinite) expression against the last response received, counting the number of results.                                      |
+|all json path matches <expression>             |Evaluates the expression against the last response received and returns all results. Intended to be used with 'show' command.                  |
+|set json path <expression> to <value> in <json>|Evaluates the expression against the provided json and -if exactly one key is found- replaces that key's existing value with the provided value|
```
fhoeben: Have you considered changing the order of 'to' and 'in'?
To me `set json path <expression> in <json> to <value>` seems more natural.

Furthermore the special 'response' value seems to warrant mentioning in this page.

tcnh: I considered the ordering to be the other way around but didn't, because a large json string would make the test log less readable when the value to set would appear last. 

Your suggestion to update the current response would void that argument so I guess that's a good idea. I'll update tonight. Should we remove the method that allows a custom string to be sent as the json to update? Can't really think of a proper use case for that, other than creating test data templates (which we can already achieve by using ftl's.)

fhoeben: If you don't have a need for that method: let's remove it.

This method might in future be better placed in a separate Json utility fixture, if it is needed.

fhoeben: Can you give a brief outline of the use case you use this feature (to alter/set a value in a json response)? All I can think of really is creation of test data (which, like you said, can also be achieved using a template). In which situations do you use this?

tcnh: In my case, we had a flow where the front-end gets a viewObject from the rest-service. The user uses a form in the front-end to update this data and post it to a json endpoint. 

In order to test back end field validations, we want to update the json and send it back, so that we can create test cases for the rest service without being dependent on the front-end.

In most cases it'l be possible to predefine the messages, but this approach leaves less clutter in the tests and is less sensitive to small changes in the view-object. 
In this case however, we needed to update a field with a value that depended on the previous value (version +1). In that case it's convenient to jus be able to update the reponse and use the updated response as the next request body.

fhoeben:
Updating a version values sounds familiar. Do you also have a way (special fixture?) to calculate the new +1 value. Or can your tests just use/assume a fixed new version?

src/main/java/nl/hsac/fitnesse/fixture/slim/JsonHttpTest.java
```diff
@@ -66,6 +66,17 @@ public String allJsonPathMatches(String expr) {
...
 
+    /**
+     * Update a value in a the response by supplied jsonPath
+     * @param path the jsonPath to locate the key whose value needs changing
+     * @param value the new value to set
+     */
+    public void setJsonPathTo(String path, String value){
+        String jsonStr = getResponse().getResponse();
+        String jsonPath = getPathExpr(path);
+        getResponse().setResponse(pathHelper.updateJsonPathWithValue(jsonStr, jsonPath, value));
```
fhoeben: Please make a local variable with the new response, before setting it. That makes debugging just so much easier...

src/main/java/nl/hsac/fitnesse/fixture/util/JsonPathHelper.java
```diff
@@ -54,4 +51,12 @@ public Object getJsonPath(String json, String jsonPath) {
         }
         return result;
     }
+
+    public String updateJsonPathWithValue(String json, String jsonPath, String value) {
+        if(null != getJsonPath(json, jsonPath)) {
```
fhoeben: This means you can't set a value for a field not previously present. Is that a requirement/good idea?
This may be an acceptable limitation for now, but just want to check.

src/main/java/nl/hsac/fitnesse/fixture/slim/JsonHttpTest.java
tcnh: Pushed.
I have a utility method for that, but it's pretty specific for this project (versions are letters, in an excel-like range A to Z, then AA to ZZ, etc.). I have it in a custom fixture class extending from jsonHttpTest in this case.

I think this kind of thing will most of the time be very project-specific and should be in an custom fixture class..

src/main/java/nl/hsac/fitnesse/fixture/util/JsonPathHelper.java
tcnh: That's intended behaviour for now (in my case, I don't need it and I think it would encourage to build fragile tests.. theoretically you could then get any arbitrary json and abuse the fixture to create something entirely different. Would not like to debug that). I didn't include an option to delete objects either :-)

Issue #65, opened 2016-06-23
Browser test action for double click


binkley:
Selenium supports sending double-click to the driver (`Actions::doubleClick`), but I did not see a way to do that with the `BrowserTest` fixture.  Did I miss something?

fhoeben:
You did not miss it, it's indeed not there (yet). 
I've looked at actions a couple of times in the past (mostly drag 'n drop related) but I have not had a need for them yet. 
If you do need double clicking, or other actions: please submit a pull request.

PR #66, opened 2016-06-24
Support double click

binkley:
Hi,

UPDATE: Simplified by leveraging `@WaitUntil`, as suggested.

UPDATE: I just recreated the pull with a documentation line in `HsacExamples.SlimTests.BrowserTests`.

I've updated the pull request for #65.  It limits implementation to just "double click", and the wiki acceptance test follows your recommendation of setting an input on a button double click.

I'm new to web testing (well ... 12 years out of practice, which amounts to the same thing), so appreciate the advice on best approach.

Cheers,
--binkley


src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -643,6 +654,21 @@ protected boolean clickImp(String place) {
...
 
+    protected boolean doubleClickImp(String place) {
+        boolean result = false;
+        try {
+            WebElement element = getElementToClick(place);
+            result = doubleClickElement(element);
+        } catch (WebDriverException e) {
+            // if other element hides the element (in Chrome) an exception is thrown
+            String msg = e.getMessage();
+            if (msg == null || !msg.contains("Other element would receive the double click")) {
```
fhoeben: Did you see the 'Other element would receive the double click' message in your tests being thrown by Chrome (or did you just copy `clickImpl()` and changed the text?

binkley:
I copied over the existing impl & changed "click" to "double click" in the message text.

Thinking this through, I'm considering "Rule of 3" -- if adding a 3rd Action needing a copy of the impl code, refactor to common, shared code.

Would you be interested in a patch which provides most of the rest of Selenium's Action class?  I'm thinking about completeness.  Contrawise, if it isn't used ... why code it?

fhoeben:
I believe we should keep things simple and minimal, until it is needed.
I would suggest:
- update the test page (simplify it) to actually (and only) see BrowserTest perform the double click (e.g. by having a javascript function on `ondblclick` that changes an input's value and see that the value changes)
- remove `doubleClickIfAvailable()` (unless you need it), and simplify the `doubleClick()` to just do the double click and not catch any exception, or worry about their message. (The `clickIfAvailable` method was added to ensure we could click a 'Logout' button at the start of a test if previous tests could leave the browser in a `logged in` state. I doubt there is a real use case for similar behavior with double clicking.)

Adding more `Action` functionality can always be done when/if it is needed. Or if there is a clear usage in a testing scenario that would be good to support (so work the other way around: what relevant end-user functionality do we want to be able to test). The only thing I thought of in the past was drag and drop, but I have not actually needed to test an application that supported that. Do you see any functionality that should be added?

wiki/FitNesseRoot/HsacAcceptanceTests/SlimTests/BrowserTest/DoubleClickByLabelOrButton/content.txt
```diff
@@ -0,0 +1,32 @@
+This test ensures that we can double click on buttons, based on their displayed text, and on labels, even if the associated input is invisible.
```
fhoeben: I believe the test now (correctly) only checks we can double click. Let's have it state that

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -643,6 +644,22 @@ protected boolean clickImp(String place) {
         return result;
     }
 
+    @WaitUntil
+    public boolean doubleClick(final String place) {
+        boolean result = false;
+        try {
+            WebElement element = getElementToClick(place);
+            result = doubleClickElement(element);
+        } catch (WebDriverException e) {
```
fhoeben: I believe this whole catch block can be removed. The standard `WaitUntil` handling will be sufficient I believe.

binkley:
Thanks for merging.  This issue was helpful for me to understand the style/preferences of the project.  I'm happy to make future contributions.  Cheers.

fhoeben:
Thanks for contributing!

Issue #63, opened 2016-06-02
Feat: Chrome Emulator


teunisnl:
For mobile website testing (without grids) this would be a nice addition
https://sites.google.com/a/chromium.org/chromedriver/mobile-emulation

in case of you prefer a pull request:
I have quick solution working (1 device) looked at converting this into SeleniumDriverSetup.java
but run out ideas how todo this clean within the current functions.
Could introduce a startChromeEmulator ? or any other ideas

fhoeben:
I have not had the time to look at this yet, sorry.
Can you share (in a fork) your current 'quick solution'?

teunisnl:
https://github.com/teunisnl/hsac-fitnesse-fixtures/tree/ChromeEmulation

no ability to define your own device just only the default list, only way to keep in same line with start driver. but i could add more flexibility to have it more configurable, but as result harder to understand

fhoeben:
I created a branch with a slightly updated version of your idea. It makes it more explicit the profile is (only) for mobile emulation. What do you think of that?

teunisnl:
well that great now it can do with defined devices and custom ones

fhoeben:
Good enough for your needs? I have spent no effort yet to think about/check how this can be used in build server job where Selenium configuration is done outside wiki.
But maybe that's not that important for these jobs (multiple wiki configs are desirable/acceptable)?

teunisnl:
More then enough. 
Seeing its not 100% guaranteed that test that work on desktop will work on mobile, so we use a 2nd suite and symbolic links to cases that are the same for mobile and desktop and make separated for those that don't. 
we are looking into it if we can use normal selenium grid and the same chrome emulator for mobile testing.
For more "realistic" mobile testing we usually use commercial grids

teunisnl:
'''
|script           |map fixture                      |
|set value        |!-Apple iPhone 6-!|for|deviceName|
|$mobileEmulation=|copy map                         |

|script         |map fixture                         |
|set value      |$mobileEmulation|for|mobileEmulation|
|$chromeoptions=|copy map                            |

|script        |map fixture                     |
|set value     |chrome        |for|browserName  |
|set value     |$chromeoptions|for|chromeOptions|
|$capabilities=|copy map                        |

|script              |selenium driver setup                                                                                                                         |
|connect to driver at|${GRID_HUB}         |with capabilities|$capabilities                                                                                          |
|show                |driver description                                                                                                                            |
'''

[...]

PR #67, opened 2016-06-27
seleniumJsonCapabilities 

teunisnl:
#63 

Tried the included jsonhelper first, but that has issues with maps within maps
Not sure if you willing to use/include Gson, otherwise I have to write a parser.
Unfortunately you will have to escape all ", or add a replace ' by " inside the new Factory

Examples that I tested (launch the chrome correct)
```
mvn clean test-compile failsafe:integration-test -DseleniumGridUrl=http://localhost:4444/wd/hub -DseleniumJsonCapabilities="{\"browserName\":\"chrome\",\"chromeOptions\":{\"mobileEmulation\":{\"deviceName\":\"Google Nexus 5\"}}}" -DfitnesseSuiteToRun=HsacExamples.SlimTests.BrowserTests.ScriptTest

or

mvn clean test-compile failsafe:integration-test -DseleniumGridUrl=http://localhost:4444/wd/hub -DseleniumJsonCapabilities="{\"browserName\":\"chrome\",\"chromeOptions\":{\"mobileEmulation\":{\"deviceMetrics\":{\"pixelRatio\":3.0,\"width\":360,\"height\":640},\"userAgent\":\"Mozilla/5.0 (Linux; Android 4.2.1; en-us; Nexus 5 Build/JOP40D) AppleWebKit/535.19 (KHTML, like Gecko) Chrome/18.0.1025.166 Mobile Safari/535.19\"}}}" -DfitnesseSuiteToRun=HsacExamples.SlimTests.BrowserTests.ScriptTest
```
ps the map fixture you suggested doesn't work, chromeOptions is really picky on what you have supply

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/junit/selenium/SeleniumJsonGridDriverFactoryFactory.java
+++ b/src/main/java/nl/hsac/fitnesse/junit/selenium/SeleniumJsonGridDriverFactoryFactory.java
@@ -1,16 +1,16 @@
 package nl.hsac.fitnesse.junit.selenium;
 
+import com.google.gson.Gson;
 import nl.hsac.fitnesse.fixture.slim.web.SeleniumDriverSetup;
 import nl.hsac.fitnesse.fixture.util.selenium.SeleniumHelper;
-import nl.hsac.fitnesse.fixture.Environment;
 
 import java.net.MalformedURLException;
-import java.util.LinkedHashMap;
+import java.util.HashMap;
 import java.util.Map;
 
 /**
  * Creates a Selenium driver factory to override the configuration in the wiki.
- * This factory is configured by setting the system property 'seleniumGridUrl' AND 'seleniumCapabilities'.
+ * This factory is configured by setting the system property 'seleniumGridUrl' AND 'seleniumJsonCapabilities'.
  */
 public class SeleniumJsonGridDriverFactoryFactory extends SeleniumDriverFactoryFactoryBase {
     @Override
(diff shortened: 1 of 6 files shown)


src/main/java/nl/hsac/fitnesse/junit/selenium/SeleniumJsonGridDriverFactoryFactory.java
```diff
@@ -0,0 +1,55 @@
...
+    protected Map<String, Object> getCapabilities() {
+        String capabilitiesString = getProperty(SELENIUM_JSONCAPABILITIES);
+        try {
+            Map<String, Object> result = new HashMap<String, Object>();
+            Gson gson = new Gson();
+            result = (Map<String, Object>)gson.fromJson(capabilitiesString, result.getClass());
+            return result;
+        } catch (Exception e) {
+            throw new RuntimeException("Unable to parse Selenium capabilities: " + capabilitiesString
+                    + "\n*", e);
```
fhoeben: Why the `+ "\n*"`?

fhoeben:
What is the problem you experienced with `JsonHelper`? Is that something that must/can be addressed?

Maybe if Gson is the answer that should be used by JsonHelper, and then the factory could use that?

src/main/java/nl/hsac/fitnesse/junit/selenium/SeleniumJsonGridDriverFactoryFactory.java
teunisnl: Left over of the Copy & Paste

teunisnl:
Issue with Jsonhelper that does not support Maps within Maps (where object is another Map<string, object>)
if I use ```|connect to driver at|${GRID_HUB}          |with json capabilities |{"browserName":"chrome", "chromeOptions":{"mobileEmulation":{"deviceName":"Google Nexus 5"}}}|``` I get an error, while this is good json param.

so fixing/changing the jsonhelper might be better option. maby there better option then gson (I couldn't find a better, less dependent solution)

teunisnl:
Not that I need it, but for acceptance testing I could add a import/export json to map fixture

fhoeben:
Your change to JsonHelper broke a unit test, I believe the new setup no longer formats one json property per line.

Furthermore it would be good to add unit test to verify the Json parser can handle nested Maps.

teunisnl:
Added unit test,
reverted back to old json for formating gson need an expected class to format

fhoeben:
I managed to make your test pass without using gson. Is that also an acceptable solution, or are the other reasons to introduce it?
See https://github.com/fhoeben/hsac-fitnesse-fixtures/tree/mobileEmulationOnGrid

teunisnl:
ill run this monday against the grid to check if chrome is still happy

teunisnl:
Oke tested it on chrome, as expected no issues.

Also tested your map fixture fix, that works too.

PR #69, opened 2016-07-04
Cookies Fix

brezelordina:


src/main/java/nl/hsac/fitnesse/fixture/util/HttpClient.java
```diff
@@ -130,17 +133,35 @@ protected void getResponse(String url, HttpResponse response, HttpRequestBase me
...
+//            } else {
+                resp = getHttpResponse(store, url, method);
+//            }
             int returnCode = resp.getStatusLine().getStatusCode();
             response.setStatusCode(returnCode);
             HttpEntity entity = resp.getEntity();
 
             Map<String, String> responseHeaders = response.getResponseHeaders();
             for (Header h : resp.getAllHeaders()) {
+            	if(h.getName().equalsIgnoreCase("Set-Cookie"))
```
fhoeben: Does this not cause too many values for 'Set-Cookie' in responseHeader when response's cookieStore was filled before request? 
I believe it will now appear as-if all cookies present in response's cookieStore were sent by server (even if they were not but added programmatically before the request was sent). That is no good. The responseHeaders should only contain those Set-Cookies actually sent by the server.


**If that is no problem** : would it not be nicer to:
- first process all headers as before. 
- And then (if `responseHeaders` contains 'Set-Cookie') put all cookies in `store` in the responseHeaders for key 'Set-Cookie'.

Basically the same as you do now, but after the for loop instead of nested inside it?

brezelordina:
Done

src/main/java/nl/hsac/fitnesse/fixture/util/HttpClient.java
```diff
@@ -142,6 +147,20 @@ protected void getResponse(String url, HttpResponse response, HttpRequestBase me
...
             }
+        	if(responseHeaders.containsKey("Set-Cookie"))
+            {
+            	StringBuilder cookieHeader = new StringBuilder();
+            	List<Cookie> cookies = store.getCookies();
+        		
+        		for(Cookie cookie : cookies)
+        		{
+        			cookieHeader.append(cookie.getName() + "=" + cookie.getValue());
+        			cookieHeader.append(";");
```
fhoeben: As far as I can tell header values should be separated by `,` not `;`. 

Still all cookies are being added to the headers (also any that were present in the store before the request). We should find a way to have only the ones from the response.

fhoeben:
I'm not sure whether merging multiple Set-Cookie headers to a single one (which is required to put it in the standard responseHeaders Map) is a good idea.
The actual response is just not something where each header name can appear only once (as you experienced with 'Set-Cookie') we should not try to work around that, or hide it like we do today, we should properly support it.

Maybe we need a new method in the fixture to check cookies (and store all the received cookies).

Or we need to change the standard map currently being used for the headers by one that supports multiple occurrences per key (but keep the existing methods that return the last value per key) and add methods to access all headers.

brezelordina:
I created the new fixture - HttpTestCS

src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTestCS.java
```diff
@@ -0,0 +1,55 @@
...
+    	if(store!=null)
+    	{
+	    	for(Cookie cookie : store.getCookies())
+	    	{
+	    		sb.append(cookie.getName() + "=" + cookie.getValue());
+	    		sb.append(";");
+	    	}
+    	}
+    	
+    	responseHeaders.put("Set-Cookie", sb.toString());
```
fhoeben: I suggest that you only read the cookie store, and put a header, if the responseHeaders contains Set-Cookie, **and the store is not null**.

brezelordina:
Done, Is it good now?

fhoeben:
I believe this will be good enough as temporary solution in your project. BUT I'm not merging it into this project in its current state. 

We should properly expose all headers received (in the order they are received) and not be merging multiple headers into a single one.

Potentially we could also offer specialized fixture methods to allow tests to access the cookies received (using the cookie store) using a nicer syntax than response headers.

fhoeben:
The features you were looking for are present in #87

PR #70, opened 2016-07-04
Allure framework

tcnh:
First attempt at allure reporting framework (http://allure.qatools.ru/) integration.
We had a situation were we needed nicer reports from fitNesse. I built an adapter/listener for allure framework that creates a nice dashboard with graphs, colors and other management-loving functionality. Use in CI builds.

In short: mini-manual/featurelist:
- The dashboard shows an aggregate of all runs in target/allure-results (so running multiple suites and generating one report is possible - don't clean between runs!)
- If tags are provided on the test pages, these are visible as functional 'stories' in the report
- HSAC's fitnesse report is copied and integrated as an attachment per test case
- If exceptions were thrown, screenshot and pagesource are attached separately

Generate results by setting createAllureResults systemproperty to true. (see HsacFitnesseRunner).
The report is generated afterwards by mvn site by default in targe/allure-report. To view the report in a browser, access this directory using a webserver (browsers won't allow XHR to file:// url's.) Fitnesse results are copied in, so you can drill down to technical results inside the dashboard.


src/main/java/nl/hsac/fitnesse/junit/JUnitAllureFrameworkListener.java
```diff
@@ -0,0 +1,266 @@
+//
+// Source code recreated from a .class file by IntelliJ IDEA
+// (powered by Fernflower decompiler)
+//
```
fhoeben: Why source generated by a decompiler?

src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
```diff
@@ -101,6 +102,9 @@ protected FitNesseContext createContext(Class<?> suiteClass) throws Exception {
     protected void runPages(List<WikiPage> pages, RunNotifier notifier) {
         boolean seleniumConfigOverridden = configureSeleniumIfNeeded();
         try {
+            if ("true".equalsIgnoreCase(System.getProperty(allureEnablerVariableName))){
```
fhoeben: This could (of course) also be done in the pom.xml (where I already add 2 listeners by default). Why does this need to happen in code (triggered by a system property)?

fhoeben:
You removed some (content of) acceptance tests. Why?

pom.xml
```diff
@@ -222,7 +230,30 @@
...
+						<phase>site</phase>
+						<goals>
+							<goal>copy-resources</goal>
+						</goals>
+						<configuration>
+							<outputDirectory>${allure.report.directory}/fitnesseResults</outputDirectory>
+							<resources>
+								<resource>
+									<directory>${project.build.directory}/fitnesse-results</directory>
+									<filtering>true</filtering>
```
fhoeben: What is the filtering used for?

fhoeben:
The changes to the projects pom are quite extensive. Would it be possible to contain these in a 'profile'? Then one could run with allure enabled something like `mvn -Pallure clean test-compile failsafe:integration-test -DfitnesseSuiteToRun=HsacExamples.SlimTests.BrowserTests.AllureReportExample`

src/main/java/nl/hsac/fitnesse/junit/JUnitAllureFrameworkListener.java
tcnh: I accidentally deleted the listener without committing. Only had the class file in the target folder, so I decompiled and refactored it back to what it was :)

src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
tcnh: We need the page tags in the listener to support grouping by tag in the report. This is now solved by instantiating the listener and giving it the pages object. if we can achieve the same by adding it to the pom, that would be the preferred way, but I don't think we can get the tags to the listener in that case?

pom.xml
tcnh: It's not. will remove it

fhoeben:
I get the impression the main bonus of using Allure is that you can create a single report over multiple runs. Correct?

In that case would it make sense to explicitly indicate this flow in the example (also highlighting the fact that clean should not be done between runs)? E.g.:
 - `mvn -Pallure clean test-compile`
 - `mvn -Pallure failsafe:integration-test -DfitnesseSuiteToRun=HsacExamples.SlimTests.BrowserTests.AllureReportExample1`
 - `mvn -Pallure failsafe:integration-test -DfitnesseSuiteToRun=HsacExamples.SlimTests.BrowserTests.AllureReportExample2`
 - ...
 - `mvn -Pallure failsafe:integration-test -DfitnesseSuiteToRun=HsacExamples.SlimTests.BrowserTests.AllureReportExampleX`
 - `mvn -Pallure site`

Having to trigger these runs one by one seems like a chore. Have you considered creating a 'parent suite' (either by using symbolic links or a search) in the wiki so all tests are run as a single (virtual) suite? In that case the current report already also has the information from all runs.

tcnh:
I'll check the acceptance tests, these should not have been changed. May be a merge/rebase issue from when I updated the changes from master into this branch.

We can also create a profile in the pom; that will, however result in some duplication in the pom, because we need the aspectJ argline to create the results-xmls. Or am I wrong?

src/main/java/nl/hsac/fitnesse/junit/JUnitAllureFrameworkListener.java
```diff
@@ -0,0 +1,266 @@
...
+        if(!this.getSuites().containsKey(suiteName)) {
+            Description suiteDescription = Description.createSuiteDescription(description.getTestClass());
+            this.testSuiteStarted(suiteDescription);
+        }
+
+        return this.getSuites().get(suiteName);
+    }
+
+    public String getIgnoredMessage(Description description) {
+        Ignore ignore = description.getAnnotation(Ignore.class);
```
fhoeben: Will this ever work for a FitNesse page? (I don't think they have `@Ignore` annotations)

fhoeben:
Using a profile will potentially introduce duplication. But maybe that can be prevented by always passing an argline property, normally it will be blank but if the profile is active the property will be filled with the aspectJ reference. 

(Can you explain why aspectJ needed anyway?)

src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
fhoeben: We could introduce a singleton (static field) to contain the list. The listener should be able to reference it from there. But that seems like a bit of a hack. 
Ideally we should be able to add the tags to the Description created for a page, that would be the preferred way to pass information. I'm not jUnit expert, but maybe we could have the runner add them as 'annotations'?

tcnh:
AspectjWeaver is used to create steps and attachments from annotations in Allure framework

src/main/java/nl/hsac/fitnesse/junit/JUnitAllureFrameworkListener.java
tcnh: No, it won't I think I missed this one when cleaning up allure's default JUnit adapter (where this adapter was based on)

tcnh:
> I get the impression the main bonus of using Allure is that you can create a single report over multiple runs. Correct?

That, and having a more apealing dashboard than a Jenkins job status page :-)

> Having to trigger these runs one by one seems like a chore. Have you considered creating a 'parent suite' (either by using symbolic links or a search) in the wiki so all tests are run as a single (virtual) suite? In that case the current report already also has the information from all runs.

Correct, but using a single unit test to run this virtual suite will cause only one testSuiteStarted/Ended event to be fired, thus reporting it as a single suite. We may be able to solve that by manually finishing the current and triggering a new testSuiteStarted event when the before-last part of the page name changes.

fhoeben:
A workaround for a project (although probably not as nice, and not so suitable to put in the baseline) to get multiple suites (and not having to do multiple maven commands) could be to create multiple test classes. Each would be annotated with `@RunWith(HsacFitNesseRunner.class)` and the suite to run is actually in the class (and therefore does not need to be supplied/overridden on the maven command line). The pom could be updated to run all of those (and not (just) FixtureDebug) classes. Each would be a different suite. And I think you could even get failsafe to run each of them in parallel (each in its own VM) to speed up overal test progress.

Even without Allure we can consider whether it would be a good idea for the runner to trigger testSuiteEnded/Started events each time the before-last part of the page name changes. That might even be a good pull request to sent to the FitNesse project for their FitNesseRunner.

But all that does not give you the nice dashboard, and that is a great feature. I think it is quite strange they need aspectJ once they have their own listener. The whole idea of the listeners is that they gather information not that you need to alter the bytecode also.

src/main/java/nl/hsac/fitnesse/junit/JUnitAllureFrameworkListener.java
```diff
@@ -0,0 +1,266 @@
...
+        String uid = this.generateSuiteUid(description.getDisplayName());
+        String suiteName = System.getProperty("fitnesseSuiteToRun");
+        if(null == suiteName) {
+            suiteName = description.getAnnotation(FitNesseRunner.Suite.class).value();
+            }
+
+        TestSuiteStartedEvent event = new TestSuiteStartedEvent(uid, suiteName);
+        AnnotationManager am = new AnnotationManager(description.getAnnotations());
+        am.update(event);
+        event.withLabels(AllureModelUtils.createTestFrameworkLabel("JUnit"));
```
fhoeben: Should this be FitNesse?

fhoeben:
Taking a quick peek at their code (https://github.com/allure-framework/allure-core/tree/master/allure-java-aspects/src/main/java/ru/yandex/qatools/allure/aspects) suggests to me aspectJ is only needed to support attachments, parameterized tests and steps defined using annotations.

I see that you use the `@Attachment` annotation. But requiring aspectJ just for that seems over the top. All the aspect the defined does is:
` Allure.LIFECYCLE.fire(new MakeAttachmentEvent(bytes, attachTitle, attachment.type()));`

Could you not just call that from the listener?

tcnh:
I'll try manually firing the attachment event. If that eliminates the need for AspectJ it's more than worth a shot :)

fhoeben:
Small sample to pass information on wiki page to any interested listener:

In HsacFitNesseRunner (or even in FitNesseRunner)
```
    @Override
    protected Description describeChild(WikiPage child) {
        String name = child.getPageCrawler().getFullPath().toString();
        FitNesseWikiPageAnnotation wikiPageAnnotation = new FitNesseWikiPageAnnotation(child);
        return Description.createTestDescription(suiteClass, name, wikiPageAnnotation);
    }
```

where the annotation is defined as:
```
/**
 * Annotation used to pass WikiPage to test listeners.
 */
public class FitNesseWikiPageAnnotation implements Annotation {
    private final WikiPage wikiPage;

    /**
     * Creates new.
     * @param wikiPage page describing test.
     */
    public FitNesseWikiPageAnnotation(WikiPage wikiPage) {
        this.wikiPage = wikiPage;
    }

    /**
     * @return wiki page for current test.
     */
    public WikiPage getWikiPage() {
        return wikiPage;
    }

    @Override
    public Class<? extends FitNesseWikiPageAnnotation> annotationType() {
        return FitNesseWikiPageAnnotation.class;
    }
}
```

With this approach any listener can get at the wiki page using:
```
        FitNesseWikiPageAnnotation pageAnn = description.getAnnotation(FitNesseWikiPageAnnotation.class);
        if (pageAnn != null) {
            WikiPage page = pageAnn.getWikiPage();
```

fhoeben:
It turns out to be a bit harder to expose the wikiPage to the listener (we need an updated version of FitNesse's JUnitRunNotifierResultsListener). In my branch I have got a working setup...

src/main/java/nl/hsac/fitnesse/junit/allure/AllureSetLabelsEvent.java
```diff
@@ -0,0 +1,25 @@
...
+import ru.yandex.qatools.allure.events.TestCaseEvent;
+import ru.yandex.qatools.allure.model.Label;
+import ru.yandex.qatools.allure.model.TestCaseResult;
+
+import java.util.List;
+
+/**
+ * Created by Tom on 4-7-2016.
+ */
+
```
fhoeben: Can you provide a more meaningful comment? ;-)

fhoeben:
You still appear to be deleting some content.txt files in the `HsacAcceptanceTests` part of the wiki. Can you address that?

tcnh:
AcceptanceTests should be back to normal now. Labels comment is a bit more descriptive, subsuites are now reported as features to have a more visual structure.

fhoeben:
I have a problem accessing accessing the fitnesse-result pages (in the allure report's iframe). Allure expects the html file to be accessible from the root of the web server (because the iframe's src starts with a '/'), that was not true for my server, so the page was not found. 
That URL should be relative to the parent page (e.g. '../fitnesseResults/HsacExamples.SlimTests.BrowserTests.ScriptTest.html' instead of '/fitnesseResults/HsacExamples.SlimTests.BrowserTests.ScriptTest.html')

Can you address this?

pom.xml
```diff
@@ -164,6 +166,11 @@
 			<version>4.12</version>
 			<scope>provided</scope>
 		</dependency>
+        <dependency>
+            <groupId>ru.yandex.qatools.allure</groupId>
+            <artifactId>allure-junit-adaptor</artifactId>
```
fhoeben: I believe `allure-java-adaptor-api` as dependency would suffice.

fhoeben:
Please add a page HsacExamples.SlimTests.AllureReportExample.SuiteTearDown that stops phantomjs. Otherwise BrowserTest examples will also try to phantom, instead of chrome, and that does not work very nicely.

fhoeben:
I added fixes to my recent comments (and some more) to https://github.com/fhoeben/hsac-fitnesse-fixtures/tree/tcnh_AllureFramework...

fhoeben:
I've updated my fork for the new 2.8.0 release.
The change to the code is now minimal, really the listener and set label event and some pom.xml changes.
Furthermore the test suite (and page property changes I cannot really explain).

But before I merge it in: can you explain to me why/how this allure report is so great? It looks a bit fancier, but I still feel like I'm clicking a lot, maybe even more than in the standard finesse-results/index.html....

tcnh:
Looks fancier, but indeed you click around more. We use it at a client that wanted a management-savvy report and the ability to easily create a per run dashboard.

In terms of information, the added value is minimal.
You could also decide to not merge it in; I'll update my fork with it then and anyone who might find it useful can get the sources from there. Same would go for the mobile testing fixtures that I suspect will get an overhaul in my current project.

bzon:
Is there any chance for this getting merged? Allure can be a great additional feature for this as an optional report. 

For example, an enterprise project that developed a backend (web services) and uses FitNesse needs to have a centralized reporting tool with a third party team that develops the front end (mobile) that uses a different testing framework.

Since most testing framework reports can be converted to [Allure](https://github.com/allure-framework/allure-core/wiki#gathering-information-about-tests), they agree that they can produce reports in their own style and have both testing preference reports to be converted in Allure in their **Integration Testing**.

fhoeben:
I don't anticipate I'll be merging this in. I can see how it can fill a need, but it can be a standalone library for those who need it. With the upcoming FitNesse release (which contains a DescriptionHelper and annotation as a needed for this) it is just two Java classes. All that would required to use it are some pom changes.

So the separate library (containing the two Java classes) could be added to an installation (as an extra dependency) and its pom updated to activate it during a test run.

I don't see the added benefit, so much, of including these classes (and the Allure jars) to this project.

fhoeben:
I've create a separate project based on this pull request: https://github.com/fhoeben/allure-fitnesse-listener. @tcnh care to add your details as developer in the pom.xml of that project?

Have any suggestions/improvements? Please supply them there, I will not incorporate this pull request into hsac-fitnesse-fixtures...

PR #76, opened 2016-08-10
Refactored chromemobileoptions to support all options types

tcnh:
This PR enables setting chromeOptions (including the ones that require a list) by using a profile.
Example: we can now pass command line arguments to a local chrome browser by starting the driver using:

| start driver for | chrome | with profile | !{args:!-[--disable-extensions, --start-maximized]-!} |

This will provide a maximized chrome window, with extensions disabled.
Without this PR, this would throw an error, because the list would be passed to DesiredCapabilities as a single String.

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/web/SeleniumDriverSetup.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/web/SeleniumDriverSetup.java
@@ -25,7 +25,9 @@ import org.openqa.selenium.safari.SafariDriver;
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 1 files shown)


fhoeben:
I'm a bit hesitant to accept this pull request. I see why it might be useful, but it's really a change on FitNesse's standard handling/syntax of Hashtables (which always produces a hash of string keys to string values).  Have you considered creating a pull request on their `fitnesse.wikitext.parser.HashTable`?

The same result can also be achieved (with a couple more lines, I admit) with my MapFixture:
```
|script   |map fixture                     |
|set value|--disable-extensions|for|args[0]|
|set value|--start-maximized   |for|args[1]|
|$options=|copy map                        |

|script          |selenium driver setup           |
|start driver for|chrome |with profile |$options  |
``` 

P.S. In your specific example I would recommend the usage of BrowserTest's (cross-browser) command `set browser size to maximum` to maximize the browser's window.

tcnh:
Didn't think of using map fixture. That's a perfectly workable solution. Let's close this PR.

PR #78, opened 2016-08-24
Response time measurement in HttpTest

fhoeben:
Adds request handling time to Http Test. And new utility fixture to measure elapsed time (and output current time) during test: `TimerFixture`

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
@@ -424,6 +424,13 @@ public class HttpTest extends SlimFixtureWithMap {
     }
 
     /**
+     * @return response time in ms for call to service.
+     */
+    public long responseTime() {
+        return getResponse().getResponseTime();
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


teunisnl:
Would be nice to introduce a Timeboard as well (based of storyboard, just times instead of screenshot)
to some kind of time measurement would be possible

fhoeben:
I thought about a 'storyboard' like table type. That would be a new feature to the hsac-fitnesse-plugin.
It would just report times, you could not make assertions based on that. I doubted how useful that would be. When 'debugging' you can just see how fast things are, in a regression test (unattended) I would always want assertions so you get a 'green or red' result

teunisnl:
It could provide a historical view of test times. (regardless of fact that browser test is not a performance test)
How useful this information is indeed the question ?
But it could provide some insight if suddenly a case/step takes longer then historic based time samples

fhoeben:
When running on a build server it already captures the time per page. How important is it to also see the times per step?

If we decided to build it: What data would you like to capture per line (e.g. current system time, time since previous step)? What format would you like to see the data?

PR #95, opened 2016-10-31
Fail build in case of test failures

daveychu:


fhoeben:
It is a good idea to document how one can make the build fail in case of a test failure. 

But that is not always what one wants: build servers may have different states for failing builds vs builds with test errors (e.g. Jenkins: FAILURE (red), UNSTABLE (blue)). In that case I prefer not to fail the build (giving it status FAILURE), but just have the failed tests recorded (giving the build status UNSTABLE).

Can you add an extra sentence, instead of just changing the command, stating that if one wants the build to fail when test failures occur one can append `...` to the command?

Issue #97, opened 2016-11-09
Website testing using custom subclass of NgBrowserTest


danielmenezesbr:
Website testing using custom subclass of NgBrowserTest doesn't work fine when a method call open(...)

Reproduce:

 - Create custom subclass of NgBrowserTest:
```
import nl.hsac.fitnesse.fixture.slim.web.NgBrowserTest;

public class CalcTest extends NgBrowserTest {
	// this method call open(...)
	public void accessMainScreen() {
		open("http://juliemr.github.io/protractor-demo/");
	}
	
	public void fillFirst(String value) {
		enter(value, "css=input[ng-model='first']", true);
		System.out.println("fillFirst");
	}
}

```
 - Go http://localhost:8080/HsacExamples.SlimTests.BrowserTests

 - Add > Test Page

 - Content:
```
!path   C:\eclipse\workspace\myapp\target\test-classes\
|script|Calc Test                          |
|access Main Screen  |
|fill First|50  |
|show |take screenshot       |foto1   |
```

 - Click "Save" button and click "Test".

 - Result:
It's produces a error: "angular could not be found on the window.". See
![hsac_result](https://cloud.githubusercontent.com/assets/5949657/20140258/5c6f9fd0-a671-11e6-9053-31e6e4b0cd2f.png)
![hsac_log](https://cloud.githubusercontent.com/assets/5949657/20140259/5c70bfe6-a671-11e6-8054-42124d41d0cf.png)

PS1: If i replace     |access Main Screen  |      by     |open|http://juliemr.github.io/protractor-demo/  | it's works fine:
![image](https://cloud.githubusercontent.com/assets/5949657/20140357/dd6b5458-a671-11e6-8e33-efbe89c0dfde.png)

fhoeben:
NgBrowserTest does not wait for Angular before it executes `open()` (since the page is not yet loaded). Your method requires similar handling. You can achieve this by overriding `requiresWaitForAngular()` in your subclass. In your override you can check whether 'accessMainScreen' is being called and return `false` if so, otherwise return the value of the superclass' implementation:
```
    @Override
    protected boolean requiresWaitForAngular(Method method) {
        if ("accessMainScreen".equals(method.getName())) {
            return false;
        } else {
            return super.requiresWaitForAngular(method);
        }
    }

```

danielmenezesbr:
@fhoeben, thanks!
It's works fine!

danielmenezesbr:
[...]

PR #102, opened 2016-11-17
Annotation @NgMethodNoWait

danielmenezesbr:
PR for annotation @NgMethodNoWait.
See #97


src/main/java/nl/hsac/fitnesse/fixture/slim/web/NgBrowserTest.java
```diff
@@ -93,7 +96,7 @@ protected void waitForAngularIfNeeded(Method method) {
      */
     protected boolean requiresWaitForAngular(Method method) {
         String methodName = method.getName();
-        return !METHODS_NO_WAIT.contains(methodName);
+        return !methodsNoWait.contains(methodName);
```
fhoeben: Why does do all methods have to be inspected in the constructor, instead of checking for the annotation in this method, after checking the static list (possibly caching the result in a field so it does not have to be repeated on each call)? That way not all methods have to be checked for each instance created, which is quite expensive time wise. And I expect most methods will not be annotated. Furthermore if the annotation is NOT present a wait was required, so the time spent checking for it is most likely not that important.

src/main/java/nl/hsac/fitnesse/fixture/slim/web/annotation/NgMethodNoWait.java
```diff
@@ -0,0 +1,9 @@
+package nl.hsac.fitnesse.fixture.slim.web.annotation;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
+public @interface NgMethodNoWait {
```
fhoeben: add `@Target(ElementType.METHOD)`

Would `NoNgWait` be a better name?

src/main/java/nl/hsac/fitnesse/slim/interaction/ReflectionHelper.java
```diff
@@ -30,4 +31,19 @@ private ReflectionHelper() {}
         }
         return result;
     }
+	
+	public static List<Method> getMethodsAnnotatedWith(final Class<?> type, final Class<? extends Annotation> annotation) {
+	    final List<Method> methods = new ArrayList<Method>();
+	    Class<?> klass = type;
+	    while (klass != Object.class) {
+	        final List<Method> allMethods = new ArrayList<Method>(Arrays.asList(klass.getDeclaredMethods()));       
+	        for (final Method method : allMethods) {
```
fhoeben: Why not iterate over `klass.getDeclaredMethods()` directly?

danielmenezesbr:
@fhoeben I adjusted the implementation.
Could you evaluate again?

src/main/java/nl/hsac/fitnesse/fixture/slim/web/NgBrowserTest.java
```diff
@@ -1,301 +1,312 @@
...
+
+    /**
+     * Determines whether method requires waiting for all Angular requests to finish
+     * before it is invoked.
+     * @param method method to be invoked.
+     * @return true, if waiting for Angular is required, false otherwise.
+     */
+    protected boolean requiresWaitForAngular(Method method) {
+        String methodName = method.getName();
+        return !METHODS_NO_WAIT.contains(methodName) && !hasMethodAnnotatedWithNgNoWait(methodName);
```
fhoeben: Why check on the method name? You have the method available: just pass the `method` so the called method can check for the annotation on that specific method.

src/main/java/nl/hsac/fitnesse/fixture/slim/web/NgBrowserTest.java
```diff
@@ -1,301 +1,312 @@
...
+    }
+
+    private boolean hasMethodAnnotatedWithNgNoWait(String methodName) {
```
fhoeben: Make this method work on the method (not its name) and check whether that has the annotation (no need to loop over all methods). And rename the method to `hasNoWaitAnnotation`

fhoeben:
I actually liked the protected method to check for the annotation being present. This allows subclasses to change/extend that part of the fixture's functionality.

Can you see whether you undo the whitespace change that causes to entire classes to be shown as changed, even though you changed a small part or even nothing in them?

danielmenezesbr:
@fhoeben please, check again: https://github.com/fhoeben/hsac-fitnesse-fixtures/pull/102/files

I fixed whitespaces and add the method protected hasNoWaitAnnotation

danielmenezesbr:
@fhoeben
What do you think will replace METHODS_NO_WAIT [1] by @NoNgWait?

[1] https://github.com/fhoeben/hsac-fitnesse-fixtures/blob/2.10.1/src/main/java/nl/hsac/fitnesse/fixture/slim/web/NgBrowserTest.java#L25

fhoeben:
I don't think that is a very good idea. I would like to keep the knowledge about Angular contained to NgBrowserTest as much as possible (even though there already is some leakage to BrowserTest). So adding annotations specifically for Angular to methods in BrowserTest does not appeal to me. I prefer to keep that to NgBrowserTest.
For methods defined by NgBrowserTest it would be an option.

danielmenezesbr:
The idea would be keep BrowserTest like as is and to do:
a) Override all methods not present in NgBrowserTest (and present in METHODS_NO_WAIT), for example:
	@Override
	@NoNgWait
	public String takeScreenshot(String basename) {
		return super.takeScreenshot(basename);
	}

b) Add @NoNgWait for all methods present in NgBrowserTest (and present in METHODS_NO_WAIT) .

This way METHODS_NO_WAIT would not be more necessary.

fhoeben:
I think the amount of source we would add to NgBrowserTest (all overrides) would be MUCH larger than justified by removing the complexity of having 2 mechanisms to ignore methods. But the approach you suggest can, technically, work.
If you have the time you could make an attempt (in a new branch) and see what the result is...

PR #103, opened 2016-11-18
Stringfixture valueOF and BrowserTest valueOf have same name

teunisnl:
Reason:
I use StringFixture and BrowserTest in combination 
```
|library|
|string fixture|
|browser test|
```
If i wanna use StringFixtures value of, i would need to change too string fixture and after that back too browser test

Propossal:
Introduce same function of valueOf under a different name, as I did in this pull request 
(this allows you to access StringFixtures valueOf by using assignTo)
Idealy would be to @Deprecated valueOf in StringFixture. but I left that open since that might have a major impact

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/StringFixture.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/StringFixture.java
@@ -18,6 +18,15 @@ public class StringFixture extends SlimFixture {
     public String valueOf(String value) {
         return value;
     }
+	
+	/**
+     * Returns value.
+     * @param value value to return
+     * @return value
+     */
+    public String assignTo(String value) {
+        return value;
+    }
 
     /**
      * Determines length of string.


fhoeben:
Can you elaborate why StringFixture's `valueOf` (or `assignTo`) is so helpful inside a browser test? I assume you are assigning a value to a variable? But the value for this variable is not know before you start working with browser test, but is also not the output of a browser test method?

teunisnl:
At the moment I using valueOf to store a !randomString 
(not the best method for filling in an unique identifier, but good enough for now)

fhoeben:
why not do that *before* you start a browser test script? Makes the test easier to read/maintain, in my opinion, as you make it clear your test uses a random value as input and this is not something that comes from the application...

Usage of 'libraries' has a tendency to encourage hard to grasp test by magically pulling in functionality.

If you really feel this is the best way forward, make your own subclass of stringfixture (with an `assignTo` method) and use that as a library. I don't see this is really a good improvement of string fixture in general.

PR #106, opened 2016-12-21
File fixture get first line

tcnh:
Added takeFirstLineFrom file method. Allows for using a file based list of non-reusable data (i.e. card numbers in a test environment) to use in your tests. May be useful for others as well

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/FileFixture.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/FileFixture.java
@@ -7,6 +7,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.util.Scanner;
 
 /**
  * Utility fixture to work with files.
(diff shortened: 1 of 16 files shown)


tcnh:
I don't understand where the old (july commits come from.. they seem not to have affected any files?
Only affected files should be FileFixture.java and the example/test wiki page.

fhoeben:
In the files changed I only see the changes to the 2 files you mention, so the rest is probably some Git branch magic affecting you.

src/main/java/nl/hsac/fitnesse/fixture/slim/FileFixture.java
```diff
@@ -49,6 +50,28 @@ public String textIn(String filename) throws IOException {
...
+        ensureParentExists(fullName);
+        File file = new File(fullName);
+        Scanner fileScanner = new Scanner(file);
+        if(fileScanner.hasNextLine()) {
+            result = fileScanner.nextLine();
+        } else{
+            throw new IOException(fullName + " is an empty file.");
+        }
+
+        //Overwrite file, minus the first line.
```
fhoeben: Any chance you are willing to rewrite this to a (streaming) write to another file (e.g. .new) and at the end remove old file and rename new file to old name?
That way the entire file does not need to be loaded in memory and there is no chance of corrupting the existing file.

Maybe a bit overkill for the purpose at hand, but that just feels safer...

tcnh: See commit below. Added a utility method to do an incremental write from the scanner using a printwriter.
This method is used to create a temporary file. Temp file gets copied over the original file and gets deleted afterwards

Is that what you meant?

PR #108, opened 2017-02-07
Added support for filtering suites by in- and excluding tags

tcnh:
Added possibility to in- and exclude tags from a command-line run.
This can be done by specifying the SuiteFilter and ExcludeSuiteFilter annotations on FixtureDebugTest, or by specifying system properties 'suiteFilter' and/or 'excludeSuiteFilter'.

The ExcludeSuiteFilter interface and getExcludeSuiteFilter method for excluding tags could also be implemented fitnesse-side IMO. They are in HsacFitnesseRunner now, so that we can use a system property.

I have kept the syntax exactly as it is in FitNesseRunner for the SuiteFilter (that one does accept a sysprop value as it is in FitNesse)

An example suite with tagged tests is provided in HsacExamples.TaggedTestcases

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
+++ b/src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
@@ -47,6 +47,9 @@ public class HsacFitNesseRunner extends FitNesseRunner {
     public final static String FITNESSE_RESULTS_PATH = "target/fitnesse-results";
     /** Property to override suite to run */
     public final static String SUITE_OVERRIDE_VARIABLE_NAME = "fitnesseSuiteToRun";
+    public final static String SUITE_FILTER_STRATEGY_OVERRIDE_VARIABLE_NAME = "suiteFilterStrategy";
+    public final static String SUITE_FILTER_OVERRIDE_VARIABLE_NAME = "suiteFilter";
+    public final static String EXCLUDE_SUITE_FILTER_OVERRIDE_VARIABLE_NAME = "excludeSuiteFilter";
     private final static String SELENIUM_DEFAULT_TIMEOUT_PROP = "seleniumDefaultTimeout";
     protected final List<SeleniumDriverFactoryFactory> factoryFactories = new ArrayList<>();
 
@@ -140,31 +143,48 @@ public class HsacFitNesseRunner extends FitNesseRunner {
 
     }
 
-    // In the original runner class, we cannot use a system property as annotation value,
-    // so we override it to work exactly like the suiteFilter annotation.
-    // This also needs the ExcludeSuiteFilter interface
+    @Override
+    protected boolean getSuiteFilterAndStrategy(Class<?> klass) throws Exception {
+        String strategy = System.getProperty(SUITE_FILTER_STRATEGY_OVERRIDE_VARIABLE_NAME);
+        if(StringUtils.isEmpty(strategy)){
+            return super.getSuiteFilterAndStrategy(klass);
+        } else if (strategy.equalsIgnoreCase("and")){
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    @Override
+    protected String getSuiteFilter(Class<?> klass) throws Exception {
+        String suiteFilter = System.getProperty(SUITE_FILTER_OVERRIDE_VARIABLE_NAME);
+        if (StringUtils.isEmpty(suiteFilter)) {
+            SuiteFilter suiteFilterAnnotation = klass.getAnnotation(SuiteFilter.class);
+            if (suiteFilterAnnotation == null) {
+                return null;
+            }
+            suiteFilter = suiteFilterAnnotation.value();
+        }
+        return suiteFilter;
+    }
+
     @Override
     protected String getExcludeSuiteFilter(Class<?> klass) throws Exception {
-        ExcludeSuiteFilter excludeSuiteFilterAnnotation = (ExcludeSuiteFilter)klass.getAnnotation(ExcludeSuiteFilter.class);
-        if (excludeSuiteFilterAnnotation == null) {
-            return null;
-        } else if (!"".equals(excludeSuiteFilterAnnotation.value())) {
-            return excludeSuiteFilterAnnotation.value();
-        } else if (!"".equals(excludeSuiteFilterAnnotation.systemProperty())) {
-            return System.getProperty(excludeSuiteFilterAnnotation.systemProperty());
-        } else {
-            throw new InitializationError("In annotation @ExcludeSuiteFilter you have to specify either \'value\' or \'systemProperty\'");
+        String excludeSuiteFilter = System.getProperty(EXCLUDE_SUITE_FILTER_OVERRIDE_VARIABLE_NAME);
+        if (StringUtils.isEmpty(excludeSuiteFilter)) {
+            ExcludeSuiteFilter excludeSuiteFilterAnnotation = klass.getAnnotation(ExcludeSuiteFilter.class);
+            if (excludeSuiteFilterAnnotation == null) {
+                return null;
+            }
+            excludeSuiteFilter = excludeSuiteFilterAnnotation.value();
         }
+        return excludeSuiteFilter;
     }
 
     @Retention(RetentionPolicy.RUNTIME)
     @Target({ElementType.TYPE})
     public @interface ExcludeSuiteFilter {
         String value() default "";
-
-        String systemProperty() default "";
-
-        boolean andStrategy() default false;
     }
 
     /**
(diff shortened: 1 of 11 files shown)


src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
```diff
@@ -136,6 +140,33 @@ protected void runPages(List<WikiPage> pages, RunNotifier notifier) {
 
     }
 
+    // In the original runner class, we cannot use a system property as annotation value,
+    // so we override it to work exactly like the suiteFilter annotation.
+    // This also needs the ExcludeSuiteFilter interface
```
fhoeben: 'interface'? I suppose you mean the 'annotation' (i.e. `@interface`)?

wiki/FitNesseRoot/HsacExamples/TaggedTestcases.wiki
```diff
@@ -0,0 +1,23 @@
...
+
+Running this suite using: 
+ * mvn test-compile failsafe:integration-test -DfitnesseSuiteToRun=HsacExamples.TaggedTestcases -DsuiteFilter=runMe -DexcludeSuiteFilter=dontRun
+  * Results in 3 test cases 001, 002 to be executed (all cases that have the tag 'runMe', and don't have the tag 'dontRun')
+ * mvn test-compile failsafe:integration-test -DfitnesseSuiteToRun=HsacExamples.TaggedTestcases -DexcludeSuiteFilter=runMe
+  * Results in 3 test cases 003, 004, 006 to be executed (all cases that don't have the tag 'runMe') 
+ * mvn test-compile failsafe:integration-test -DfitnesseSuiteToRun=HsacExamples.TaggedTestcases -DsuiteFilter=runMe
+  * Results in 3 test cases 001, 002, 005 to be executed (all cases that do have the tag 'runMe')
+  
+Passing a comma separated list to the (exlude)suiteFilter is allowed, so we can also use:
```
fhoeben: Is there a way to do an 'AND', instead of the 'OR' provided by the comma?

src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
```diff
@@ -136,6 +140,33 @@ protected void runPages(List<WikiPage> pages, RunNotifier notifier) {
...
+    @Override
+    protected String getExcludeSuiteFilter(Class<?> klass) throws Exception {
+        ExcludeSuiteFilter excludeSuiteFilterAnnotation = (ExcludeSuiteFilter)klass.getAnnotation(ExcludeSuiteFilter.class);
+        if (excludeSuiteFilterAnnotation == null) {
```
fhoeben: Currently `HsacFitNesseRunner` makes the system property `fitnesseSuiteToRun` 'win' from the string literal in the `Suite` annotation (even if the Java test class code does not reference the system property). This allows the developer to commit any value in the test class without breaking the build server's run (which sets the system property).

The approach for `(Exclude)SuiteFilter` in this version is that the 'value' (literal `value` in Java code) wins from the system property.
Can we change the behavior for `(Exclude)SuiteFilter` so that the system properties (with names you use as examples) always win from the annotation value in code (even let the filters become active is the annotation was not present on the Java class)?
I believe all it would take is an extra (first) `if`-branch, turning the current first into an `else if`, checking whether the system property has a value and using that value if so.
Of course this does also mean we have to override the current behavior for the `SuiteFilter` annotation.

I also agree that the current code would be a good improvement for the FitNesse project's `FitNesseRunner`. Are you planning to send them a pull request with it?

fhoeben: I've created a pull request to FitNesse project for the exclude filter based on annotations https://github.com/unclebob/fitnesse/pull/1033

wiki/FitNesseRoot/HsacExamples/TaggedTestcases.wiki
tcnh: We can set AND/OR in fitnessedebugtest by passing the boolean andStrategy to the annotation; This now defaults to false. Would we want this to be configurable?

src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
tcnh: Should be an easy fix. I'll look into it

src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
tcnh: haha... copy/paste ftw

wiki/FitNesseRoot/HsacExamples/TaggedTestcases.wiki
fhoeben: I would like full control over the tests executed, without changing a line of Java code. So yes, I would like and vs or to be controlled by a system property also...

tcnh: I can make the suiteFilter configurable (so we can use a suiteFilterStrategy sysprop to set it to AND). This works, because FitNesse supports AND/OR for suiteFilters. For excludeSuiteFilter however, Fitnesse only supports OR. See SuiteFilter constructor:

`public SuiteFilter(String orTags, String mustNotMatchTags, String andTags, String startWithTest) {
    this.startWithTest = (!"".equals(startWithTest)) ? startWithTest : null;
    if(andTags != null) {
      matchTags = new SuiteTagMatcher(andTags, true, true);
    } else {
      matchTags = new SuiteTagMatcher(orTags, true, false);
    }
    notMatchTags = new SuiteTagMatcher(mustNotMatchTags, false, false);
  }`

The only way I see to get the excludeFilter to be configurable would be overriding most of the SuiteFilter class with a custom (HsacSuiteFilter) class, which would cause a lot of private method duplication. I'd opt to keep this filter's behaviour similar to the behaviour in Fitnesse's suite responder. (configurable AND/OR for tags to include, OR for tags to exclude). Agree?

fhoeben: I agree we should only have 'OR' for exclude filter, so we should also remove the field from the annotation to go to 'and-mode' as the framework does not support it for excludes. Have you seen the pull request I made to FitNesse? In that I also did not have 'AND' support in the exclude.

Let's keep it simple for now and only take on the extra complexity when we find we need 'AND' in our exclude filters...

src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
```diff
@@ -140,31 +143,48 @@ protected void runPages(List<WikiPage> pages, RunNotifier notifier) {
...
-            throw new InitializationError("In annotation @ExcludeSuiteFilter you have to specify either \'value\' or \'systemProperty\'");
+            return false;
+        }
+    }
+
+    @Override
+    protected String getSuiteFilter(Class<?> klass) throws Exception {
+        String suiteFilter = System.getProperty(SUITE_FILTER_OVERRIDE_VARIABLE_NAME);
+        if (StringUtils.isEmpty(suiteFilter)) {
+            SuiteFilter suiteFilterAnnotation = klass.getAnnotation(SuiteFilter.class);
```
tcnh: Somehow that doesn't work. When the SuiteFilter system property is not given and the annotation is empty, FitNesseRunner throws an initializationException. I don't really see why, but it does happen. I guess the default value for the annotation is seen as neither null or "".

src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
```diff
@@ -140,31 +143,48 @@ protected void runPages(List<WikiPage> pages, RunNotifier notifier) {
...
+            suiteFilter = suiteFilterAnnotation.value();
+        }
+        return suiteFilter;
+    }
+
+    @Override
+    protected String getExcludeSuiteFilter(Class<?> klass) throws Exception {
+        String excludeSuiteFilter = System.getProperty(EXCLUDE_SUITE_FILTER_OVERRIDE_VARIABLE_NAME);
+        if (StringUtils.isEmpty(excludeSuiteFilter)) {
+            ExcludeSuiteFilter excludeSuiteFilterAnnotation = klass.getAnnotation(ExcludeSuiteFilter.class);
```
tcnh: In this case, we can call the method from the superclass. (Simply null or value. No extra's there)

src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
```diff
@@ -140,31 +143,48 @@ protected void runPages(List<WikiPage> pages, RunNotifier notifier) {
...
+                return null;
+            }
+            excludeSuiteFilter = excludeSuiteFilterAnnotation.value();
         }
+        return excludeSuiteFilter;
     }
 
     @Retention(RetentionPolicy.RUNTIME)
     @Target({ElementType.TYPE})
     public @interface ExcludeSuiteFilter {
```
tcnh: Yes, we need it. Fitnesse's version omits the default (empty) value.

fhoeben:
@tcnh I took the liberty of making some further changes to the code in the master branch. When I run the examples it still seems to work (with less code). Did I overlook something, or is this acceptable for you?

tcnh:
If all the examples work it should be good to go :)

PR #109, opened 2017-02-15
subclass for Angular 2

greg73:
I've created a new class for Angular 2
This has the same functionality as protractor currently, i.e. no custom selectors or repeaters


fhoeben:
Looks quite good. There seems to be a bit of overlap between Ng2BrowserTest and NgBrowserTest. Could a shared superclass be defined (extracted?) for this shared functionality (e.g. the NO_WAIT methods)?

fhoeben:
Is there a public site we could create a script using this fixture for inside the example tests?

fhoeben:
@greg73 can you take a look at https://github.com/fhoeben/hsac-fitnesse-fixtures/tree/baseClassForAngularTests does that make sense for you. Does it work with Angular2 for you?

greg73:
@fhoeben base class refactoring looks good and works fine, thanks
The normal way to set up an online demo is using plunker, I'm having a look at migrating the calculator into ng2 using plunker

PR #110, opened 2017-02-16
Example test for Ng2BrowserTest

greg73:
Here's an example test for a Angular 2 port of the calculator app
As it's in plunker I had to load open the page using BrowserTest first and wait for the app to load

### DIFF
--- a/wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/Angular2CalculatorTest.wiki
+++ b/wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/Angular2CalculatorTest.wiki
@@ -5,22 +5,15 @@ Sample test on [[Angular][https://angular.io/]] application.
 !note Due to the nature of [[Plunker][https://plnkr.co/]] which hosts the application in an iframe, the page must first be opened with !-BrowserTest-!, switch to the embedded iframe and wait for the application to load.
 
 
-|table template|add                      |
-|enter         |@{first} |as |id=first   |
-|enter         |@{second}|as |id=second  |
-|select        |+        |for|id=operator|
-|click         |Go!                      |
-|$result=      |value of |id=latest      |
+|table template|add                   |
+|enter         |@{first} |as |First   |
+|enter         |@{second}|as |Second  |
+|select        |+        |for|Operator|
+|click         |Go!                   |
+|$result=      |value of |Result      |
 
-|script                |Browser Test                                |
-|seconds before timeout|15                                          |
-|set browser size to maximum                                        |
-|open                  |https://embed.plnkr.co/2GtNBG6QrG8AL0l9rxjc/|
-|switch to frame       |css=.plunker-space-iframe                   |
-|switch to frame       |id=preview                                  |
-|wait for visible      |id=calculator-container                     |
-
-|script|Ng2 Browser Test|
+|script|Ng2 Browser Test                            |
+|open  |https://greg73.github.io/angular-calculator/|
 
 |add                 |
 |first|second|result?|
--- a/wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/Angular2CalculatorTest.wiki
+++ b/wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/Angular2CalculatorTest.wiki
@@ -1,0 +1,28 @@
+Sample test on [[Angular][https://angular.io/]] application. 
+
+!note '''Please note:''' The application used in this test does NOT seem to work with !-PhantomJs-!. It does work when executed with Chrome or Firefox.
+
+!note Due to the nature of [[Plunker][https://plnkr.co/]] which hosts the application in an iframe, the page must first be opened with !-BrowserTest-!, switch to the embedded iframe and wait for the application to load.
+
+
+|table template|add                      |
+|enter         |@{first} |as |id=first   |
+|enter         |@{second}|as |id=second  |
+|select        |+        |for|id=operator|
+|click         |Go!                      |
+|$result=      |value of |id=latest      |
+
+|script                |Browser Test                                |
+|seconds before timeout|15                                          |
+|set browser size to maximum                                        |
+|open                  |https://embed.plnkr.co/2GtNBG6QrG8AL0l9rxjc/|
+|switch to frame       |css=.plunker-space-iframe                   |
+|switch to frame       |id=preview                                  |
+|wait for visible      |id=calculator-container                     |
+
+|script|Ng2 Browser Test|
+
+|add                 |
+|first|second|result?|
+|1    |2     |3      |
+|3    |4     |7      |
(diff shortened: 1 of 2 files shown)


wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/Angular2CalculatorTest.wiki
```diff
@@ -0,0 +1,28 @@
+Sample test on [[Angular][https://angular.io/]] application. 
+
+!note '''Please note:''' The application used in this test does NOT seem to work with !-PhantomJs-!. It does work when executed with Chrome or Firefox.
+
+!note Due to the nature of [[Plunker][https://plnkr.co/]] which hosts the application in an iframe, the page must first be opened with !-BrowserTest-!, switch to the embedded iframe and wait for the application to load.
```
fhoeben: The plunker page does not work for me (in Safari or chrome), maybe because the page needs files called '...js' while the plane on the left shows those files as '...ts' (i.e. with '.ts' extension)?

greg73: The service seems flaky today (giving 404 errors in the console)... try a few refreshes. (The ts is transpiled to js in the browser)

fhoeben: This seems to be quite unstable indeed. Not to great for an example then. Can we find a production site we can demo on?

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/Angular2CalculatorTest.wiki
```diff
@@ -0,0 +1,28 @@
...
+|script                |Browser Test                                |
+|seconds before timeout|15                                          |
+|set browser size to maximum                                        |
+|open                  |https://embed.plnkr.co/2GtNBG6QrG8AL0l9rxjc/|
+|switch to frame       |css=.plunker-space-iframe                   |
+|switch to frame       |id=preview                                  |
+|wait for visible      |id=calculator-container                     |
+
+|script|Ng2 Browser Test|
+
```
fhoeben: When the service was working I thought the test looked a bit better when I re-wrote the scripts to:
```
|script                |Browser Test                                |
|seconds before timeout|25                                          |
|open                  |https://embed.plnkr.co/2GtNBG6QrG8AL0l9rxjc/|
|switch to frame       |css=.plunker-space-iframe                   |
|switch to frame       |id=preview                                  |

|script          |Ng2 Browser Test|
|wait for visible|Super Calculator|

```

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/Angular2CalculatorTest.wiki
```diff
@@ -0,0 +1,28 @@
...
+!note '''Please note:''' The application used in this test does NOT seem to work with !-PhantomJs-!. It does work when executed with Chrome or Firefox.
+
+!note Due to the nature of [[Plunker][https://plnkr.co/]] which hosts the application in an iframe, the page must first be opened with !-BrowserTest-!, switch to the embedded iframe and wait for the application to load.
+
+
+|table template|add                      |
+|enter         |@{first} |as |id=first   |
+|enter         |@{second}|as |id=second  |
+|select        |+        |for|id=operator|
+|click         |Go!                      |
```
fhoeben: It would be even better if the first and second had their own placeholder and the operator an aria-label attribute. Then we could use the placeholders and the aria-label, instead of those ugly ids in the table template....

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/Angular2CalculatorTest.wiki
fhoeben: What would you think of?
```
Sample test on [[Angular][https://angular.io/]] application. 

|storyboard|ng2 browser test                        |
|open      |http://oidamo.de/angular2-image-gallery/|
|enter     |20     |as     |Space between images    |
|enter     |14     |as     |Image size              |
|press     |shift+tab                               |
|scroll to |built with                              |

```

or

```
|script|ng2 browser test                                           |
|open  |https://tahaipek.github.io/Angular-2-Github-Search-Profile/|
|enter |greg73        |as        |Enter A Github Username...       |
|click |hsac-fitnesse-fixtures                                     |

```

greg73: I like the github search one, This was the only public site I could find but it doesn't seem to have anywhere to enter text - https://code.gov

fhoeben: I just looked around a bit on http://angularexpo.com/

fhoeben: The only problem I found with both the sites I tried was that they didn't need Ng2BrowserTest, the sample tests work just as well with BrowserTest ;-)
(This is actually similar to what I found for AngularJs sites: I tend not to need NgBrowserTest and just use BrowserTest for my tests.)

Is there any particular feature you found in your own apps that Ng2BrowserTest was needed for? It would be good if we could find a sample that really relied on the specialised fixture...

fhoeben: I liked this site as well (but again ng2 browser test does not seem needed):
```
|script         |ng2 browser test                                                   |
|open           |http://www.hiteshbalar.com/preserver/notes                         |
|click          |Write a note                                                       |
|enter          |Find an Angular2 sample site     |as         |Title                |
|enter          |check out angularexpo.com        |as         |Write a note         |
|click          |Save                                                               |
|click          |Write a note                                                       |
|enter          |Find another Angular2 sample site|as         |Title                |
|enter          |check out code.gov               |as         |Write a note         |
|click          |Save                                                               |
|open in new tab|http://angularexpo.com                                             |
|show           |take screenshot                  |expo                             |
|close tab                                                                          |
|hover over     |Find an Angular2 sample site                                       |
|click          |Archive                                                            |
|ensure         |is visible                       |Find another Angular2 sample site|

```

greg73: It's needed when there is something async happening which then updates the page (usually http request, in the calculator it's just a setTimeout), then the "wait for angular" call waits for anything async to be completed. Try the calculator test with BrowserTest and it will fail,  If you want to stick with the calculator I can build a compiled version and upload to github pages?

fhoeben: Waiting for the async events is indeed why NgBrowserTest was introduced initially. Since then BrowserTest gotten better at waiting for things to be available (if you want to click/check something it will wait for it if it's not there initially), but indeed some things still need it (so far only none of the Angular2 sample sites I found). 

I prefer examples 'from the wild' but to get the sample running we could indeed place it on GitHub pages (or include it in the 'files section' of the wiki and serve via the mock server in the fixtures (i.e. 'mock xml server setup') so we can even use it in the acceptance tests). Having a good check in the acceptance test set would of course be my preferred outcome :-)

greg73: Thanks, I've improved the sample, now on github pages, and the test. I'm not familiar with how the acceptance tests work but will have another look when time allows

fhoeben:
Merged into master

Issue #130, opened 2017-05-02
Clone project is huge (4Gb)


keetron:
When cloning the project, it takes super long on my wifi and I found it is because the .pack file in the .git folder is 4Gb large. I do not feel this project has the codebase to allow for 4Gb?

fhoeben:
You are absolutely right. I had not noticed it, since I only get incremental updates, but the repo size is just too big.

I see two main factors that make big:
- The build process, which runs in Travis CI, commits build output (test results and standalone.zip) to the gh-pages branches so they are publicly available, and these are big (the zip is 60MB each build).
- Selenium web driver binaries (and their older versions) included in the wiki.

I will try to find an alternative approach to publish build results so they don't increase the size of the repo (I hoped/expected these would only be retrieved when getting that branch, but they are always included in a normal clone).

After that is sorted I can do a cleanup of the repo, removing the no longer needed files, and push this to GitHub. To benefit from this everyone will need to do a new clone, and all pull request must be (re-) based on this updated version of the repo, or else the old content will just come back and increase the size again.

tcnh:
For the time being, when you just want to clone the latest version, use the
depth parameter with git clone and set it tot a low number (i.e. 1). You
won't have the full history, but you do get a working copy.

Op 2 mei 2017 09:29 schreef "Fried Hoeben" <notifications@github.com>:

fhoeben:
I believe I have updated the build process to no longer push the standalone.zip to gh-pages on each build. I also adjusted pointers in documentation (GitHub wiki, README, and HsacExamples wiki page in project) to point to the standalone.zip in the Maven SNAPSHOT repository of Sonatype.
This should stop the growth of repo with 60MB per build.

fhoeben:
I believe the repo size is now 2.9 MB, better?

keetron:
I think that is magical.
Now to update my fork 😄

keetron:
Hmmm, how to make it so that my fork is small again?
There is a prune command I think?

fhoeben:
I believe the easiest way is to delete your fork and make a new one...

keetron:
It seems I will have to pick this up on my desktop as I get into issues when doing a fetch on my laptop. No idea why, it seems to be a network thing.

fhoeben:
I'm not sure you can do a fetch from a previous clone of the repo. Do you also have problem when doing a clone on the laptop?

keetron:
[...]

PR #136, opened 2017-05-24
Json http path exists

tcnh:
Added a method to json http test to check if json path exists in a json file.
Note: when the key exists, but value is explicitly set to null, this will return false

This PR seems huge because I branched from my fork that was pretty far behind. Only the four relevant files are changed though.


fhoeben:
Was your fork based on the updated repo version (with the smaller size #130)? I believe it is not.
I'm afraid merging this will bring back all history (and the large size). Can you create a patch containing this PR's changes and apply that on a 'clean fork'?

src/main/java/nl/hsac/fitnesse/fixture/util/JsonPathHelper.java
```diff
@@ -68,6 +68,18 @@ public String updateJsonPathWithValue(String json, String jsonPath, Object value
         }
     }
 
+    public boolean jsonPathExists(String json, String jsonPath) {
+        boolean result = false;
+        try {
+            result = null != getJsonPath(json, jsonPath);
+        } catch (RuntimeException e) {
+            if(e.getMessage().contains("multiple")) {
```
fhoeben: space after `if`, before `(`

wiki/FitNesseRoot/HsacAcceptanceTests/SlimTests/HttpTest/JsonHttpTest.wiki
```diff
@@ -31,7 +31,9 @@
 |set json path|airplanes[0].seats   |to                  |300                                                 |
 |check        |json path            |airplanes[0].seats  |300                                                 |
 |show         |response                                                                                       |
-
+|ensure       |json path exists     |airplanes[0].seats                                                       |
+|reject       |json path exists     |airplanes[0].beds                                                        |
+|ensure       |json path exists     |airplanes[*].airline                                                     |
```
fhoeben: Can you add a line
```
|reject       |json path exists     |airplanes[*].beds |
```

tcnh:
Creating new PR based on clean fork. Aso implementing suggested changes.
Nice catch with the added test btw. that situation was unhandled :)

PR #138, opened 2017-05-29
click with offset x y

teunisnl:


fhoeben:
This is quite technical, especially with the name. Furthermore how is the offset calculated? Maybe it would be better to name it something like: click <x> pixels right and <y> pixels below <place>? That's not quite right either, but....
Can you try to write the documentation for the new commands?

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -716,6 +716,29 @@ protected boolean doubleClick(WebElement element) {
         return result;
     }
 
+    @WaitUntil
+    public boolean clickWithOffsetXY(String place,int xOffset,int yOffset) {
```
fhoeben: spaces after `,`... ;-)

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -716,6 +716,29 @@ protected boolean doubleClick(WebElement element) {
...
+    @WaitUntil
+    public boolean clickWithOffsetXYIn(String place,int xOffset,int yOffset, String container) {
+        WebElement element = getElementToClick(place, container);
+        return clickWithOffsetXY(element, xOffset, yOffset);
+    }
+
+    protected boolean clickWithOffsetXY(WebElement element,int xOffset,int yOffset) {
+        boolean result = false;
+        if (element != null) {
+            scrollIfNotOnScreen(element);
```
fhoeben: Should we not scroll to the location where we are going to click? Maybe first scroll to the element, and then scroll with the offset?

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
```diff
@@ -967,6 +967,10 @@ public void doubleClick(WebElement element) {
         getActions().doubleClick(element).perform();
     }
 
+    public void clickWithOffset(WebElement element,int xOffset,int yOffset) {
+        getActions().moveToElement(element, xOffset, yOffset ).click().build().perform();
```
fhoeben: No need to call build, you can just call `perform()` after click.

I now see that the offset is relative to top left hand corner of element. That needs to be in documentation...

wiki/FitNesseRoot/HsacAcceptanceTests/SlimTests/BrowserTest/ClickWithOffset.wiki
```diff
@@ -0,0 +1,54 @@
+---
```
fhoeben: I appreciate the test!

teunisnl:
Added the documentation

on technical part, well getActions() is public now so i can always use it in an extended version

wiki/FitNesseRoot/HsacAcceptanceTests/SlimTests/BrowserTest/ClickWithOffset.wiki
teunisnl: next time make it a comment instead of a change request

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests.wiki
```diff
@@ -56,6 +56,7 @@ The heuristic to find places tries a number of translations of end user terms to
...
+|click with offset <place> X <XinPixels> Y <YinPixels>                     |clicks with a offset X,Y (offset measured in pixels of Top Left corner of element)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     |
```
fhoeben: In your test you actually use: `click <place> with offset X <horizontalOffset> Y < verticalalOffset > `
I have to say I like that better. 
Why is 'Top Left' capitalised?

PR #139, opened 2017-05-29
Drag and drop

teunisnl:
html5 does not work ([issue/3604](https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/3604))
also cant tell if that jQuery solution would work on html4 version

An accpetence test is out my skill reach; 
Also made GetActions public so i can use directly in extended versions


keetron:
Oh this is a nice to have! Not that I need it now, but if I need it, I will not have to build it myself :smile:

PR #144, opened 2017-06-13
java8 compatible and support for firefox

roydekleijn:


src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
```diff
@@ -1297,7 +1297,7 @@ public String takeScreenshot(String baseName) {
         if (t != null) {
             if (t instanceof ScreenshotException) {
                 String encodedScreenshot = ((ScreenshotException)t).getBase64EncodedScreenshot();
-                result = new Base64Encoder().decode(encodedScreenshot);
+                result = new Base64().decode(encodedScreenshot);
```
fhoeben: Let's not add a separate dependency to do base64 decode. Can't we use https://docs.oracle.com/javase/8/docs/api/java/util/Base64.Decoder.html?

pom.xml
```diff
@@ -1,12 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
fhoeben: You changed quite a few lines, and reordered some elements. This confuses the diff. Could you try to minimise your changes to what is needed, so that the diff is nice and clear?

src/test/java/nl/hsac/fitnesse/fixture/util/BsnUtilTest.java
```diff
@@ -12,7 +12,7 @@
     private final BsnUtil generator = new BsnUtil();
 
     /**
-     * Tests basic generation.
+     * Tests basic  generation.
```
fhoeben: Why an extra space?

pom.xml
```diff
@@ -280,6 +287,18 @@
 							<version>2.29</version>
 						</driver>
 						<driver>
+							<name>chromedriver</name>
+							<platform>linux</platform>
+							<bit>32</bit>
+							<version>2.29</version>
```
fhoeben: As long as we're upgrading. Can you go to chromedriver 2.30?

src/test/java/nl/hsac/fitnesse/fixture/util/BsnUtilTest.java
roydekleijn: indeed not needed... but this was needed to trigger a new build :)

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
roydekleijn: thanks, changed that

fhoeben: Don't forget to remove the extra dependency from pom.xml also....

pom.xml
```diff
@@ -262,10 +262,10 @@
 							<version>3.4.0</version>
 						</driver>
 						<driver>
-							<name>edgedriver</name>
+							<name>internetexplorerdriver</name>
```
fhoeben: Why do you change the order of the drivers? Please keep it as-is, so it is clear which versions are updated and which are added.

pom.xml
fhoeben: Please remove the linux web drivers from the project. These makes the 'standalone.zip' really huge and people running Linux tend to be technical enough to add the drivers themselves...

fhoeben:
Should we include 64-bit drivers for IE and gecko on Windows? Previously I only bundled 32-bit for IE, since that is the often used browser (even on 64-bit systems). Do you know whether both versions are needed for gecko? I prefer to bundle only the most used versions as they add quite a bit of size to the standalone.zip.

fhoeben:
Does the current code work for you? For me it doesn't (I was alerted by someone earlier that there was a issue with my Selenium usage for 3.x). When I run http://localhost:9090/HsacExamples.SlimTests.BrowserTests.AngularUiSelectTest?test (with chrome) I get an exception:
```
java.lang.NoSuchMethodError: org.openqa.selenium.support.ui.FluentWait.until(Ljava/util/function/Function;)Ljava/lang/Object;
 at nl.hsac.fitnesse.fixture.util.selenium.SeleniumHelper.waitUntil(SeleniumHelper.java:1126)
 at nl.hsac.fitnesse.fixture.slim.web.BrowserTest.waitUntilImpl(BrowserTest.java:1843)
 at nl.hsac.fitnesse.fixture.slim.web.BrowserTest.waitUntil(BrowserTest.java:1773)
 at nl.hsac.fitnesse.fixture.slim.web.BrowserTest.open(BrowserTest.java:182)
 at nl.hsac.fitnesse.fixture.slim.web.NgBrowserTestBase.open(NgBrowserTestBase.java:100)
```

This will definitely will need to be addressed :-)

fhoeben:
When I run the HsacAcceptanceTests.SlimTests.BrowserTest suite on my mac with latest firefox (from the wiki), not all tests are green. For instance the DoubleClick tests fails.

Did you run these test (on which OS), and were the green for you?

roydekleijn:
Vacation is almost over:) I will take a look in the weekend.


Verstuurd vanaf mijn iPhone

fhoeben:
I did a more thorough investigation of the issue. I believe the issues are all issues that stem from Firefox (or its web driver) that we cannot easily prevent/work around. 

For instance inability to double click or drag/n/drop are known limitations of the web driver.
The page source retrieval seems to re-order element attributes, but that should not matter in any case.
Finding links by their 'capitalised' text, when css text transform is in place does not seem to be supported (and neither is it for Safari)..

PR #147, opened 2017-07-09
most recent version of geckodriver

roydekleijn:


fhoeben:
already merged with java 8 support pull request

PR #148, opened 2017-07-11
again geckodriver update

roydekleijn:
now version 0.18


fhoeben:
Plugin does not seem to find 0.18 version yet... (see travis build)
Do we need to configure an explicit URL for that version (I believe I saw in the documentation that this is possible)

teunisnl:
Or check if its on https://github.com/webdriverextensions/webdriverextensions-maven-plugin-repository
before making the PR;

roydekleijn:
it has now been added.... I changed my archetype project so that I will download the specific drivers myself

PR #155, opened 2017-07-30
Ability to execute javascript

roydekleijn:


fhoeben:
Thanks. Looks good.

Could you:
 * look at adding a page testing this new feature to HsacAccentanceTests
 * see whether you can also allow the script parameter to be pre-formatted (i.e. remove pre-formatting inside the method, by feeding it to `cleanupValue()`)

wiki/FitNesseRoot/HsacAcceptanceTests/SlimTests/BrowserTest/ExecuteScript.wiki
```diff
@@ -0,0 +1,29 @@
...
+
+|script      |mock xml server setup|
+|add response|${HTML}              |
+|$url=       |get mock server url  |
+
+|script                |browser test                                                    |
+|seconds before timeout|1                                                               |
+|open                  |$url                                                            |
+|execute script        |document.getElementById("dateField").setAttribute("type","text")|
+|check                 |value of attribute    |type    |on    |id=dateField    |text    |
```
fhoeben: Presumably in your real tests you do an `enter <value> as id=datefield` after changing the type? Is there a single date format that must always be used, or is that browser/user locale dependent?

Would it be a good idea to add (as a separate PR) a new BrowserTest command to select a date. The java method would set the attribute type (on the element found using browser test normal heuristic, passing it as `arguments[0]`), enter the value, and change the attribute back?

PR #158, opened 2017-08-07
DateFormatFixture to do simple date format conversions

tcnh:
Simple utility fixture to convert between different date formats and translate (milli)seconds from epoch to formatted date strings and vice versa.

Also includes a wiki page with some usage examples.

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
@@ -1,0 +1,64 @@
+package nl.hsac.fitnesse.fixture.slim;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+/**
+ * Fixture that helps converting dates to and from various formats. Intended to be used as a utility fixture i.e. as library
+ */
+
+public class DateFormatFixture extends SlimFixture {
+    private String dateFormat = "dd-MM-yyyy";
+    private boolean timestampHasMilliseconds = true;
+
+    public void setDateFormat(String df) {
+        dateFormat = df;
+    }
+
+    public void timestampHasMilliseconds(boolean hasMillis) {
+        timestampHasMilliseconds = hasMillis;
+    }
+
+    public String formatTimestamp(long timestamp) {
+        return formatTimestampAs(timestamp, dateFormat);
+    }
+
+    public String formatTimestampAs(long timestamp, String dateFormat) {
+        if(!timestampHasMilliseconds) {
+            timestamp = timestamp * 1000L;
+        }
+
+        Date date = new Date(timestamp);
+        SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
+        return sdf.format(date);
+    }
+
+    public String formatDateAs(String date, String newFormat) {
+        SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
+        SimpleDateFormat targetFormat =new SimpleDateFormat(newFormat);
+        String formattedDate = null;
+        try {
+            formattedDate = targetFormat.format(sdf.parse(date));
+        } catch (ParseException e) {
+            e.printStackTrace();
+        }
+        return formattedDate;
+    }
+
+    public long timestampForDate(String date) {
+        long timeStamp = 0;
+        SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
+
+        try {
+            timeStamp = sdf.parse(date).getTime();
+            if (!timestampHasMilliseconds) {
+                timeStamp = timeStamp / 1000L;
+            }
+        } catch (ParseException e) {
+            e.printStackTrace();
+        }
+        return timeStamp;
+    }
+}
+
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
@@ -3,6 +3,7 @@ package nl.hsac.fitnesse.fixture.slim;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.TimeZone;
 
 /**
  * Fixture that helps converting dates to and from various formats. Intended to be used as a utility fixture i.e. as library
@@ -10,10 +11,15 @@ import java.util.Date;
 
 public class DateFormatFixture extends SlimFixture {
     private String dateFormat = "dd-MM-yyyy";
+    private TimeZone timezone = TimeZone.getTimeZone("GMT+1");
     private boolean timestampHasMilliseconds = true;
 
     public void setDateFormat(String df) {
         dateFormat = df;
+    }
+
+    public void setTimezone(String newTimezone) {
+        timezone = TimeZone.getTimeZone(newTimezone);
     }
 
     public void timestampHasMilliseconds(boolean hasMillis) {
@@ -31,32 +37,35 @@ public class DateFormatFixture extends SlimFixture {
 
         Date date = new Date(timestamp);
         SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
+        sdf.setTimeZone(timezone);
         return sdf.format(date);
     }
 
     public String formatDateAs(String date, String newFormat) {
         SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
+        sdf.setTimeZone(timezone);
         SimpleDateFormat targetFormat = new SimpleDateFormat(newFormat);
-        String formattedDate = null;
+        targetFormat.setTimeZone(timezone);
+        String formattedDate;
         try {
             formattedDate = targetFormat.format(sdf.parse(date));
         } catch (ParseException e) {
-            e.printStackTrace();
+            throw new SlimFixtureException(false, e.getMessage(), e);
         }
         return formattedDate;
     }
 
     public long timestampForDate(String date) {
-        long timeStamp = 0;
+        long timeStamp;
         SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
-
+        sdf.setTimeZone(timezone);
         try {
             timeStamp = sdf.parse(date).getTime();
             if (!timestampHasMilliseconds) {
                 timeStamp = timeStamp / 1000L;
             }
         } catch (ParseException e) {
-            e.printStackTrace();
+            throw new SlimFixtureException(false, e.getMessage(), e);
         }
         return timeStamp;
     }
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
@@ -43,13 +43,13 @@ public class DateFormatFixture extends SlimFixture {
         }
 
         Date date = new Date(timestamp);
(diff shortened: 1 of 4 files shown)


src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public String formatDateAs(String date, String newFormat) {
+        SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
+        SimpleDateFormat targetFormat = new SimpleDateFormat(newFormat);
+        String formattedDate = null;
+        try {
+            formattedDate = targetFormat.format(sdf.parse(date));
+        } catch (ParseException e) {
+            e.printStackTrace();
```
fhoeben: Instead of printing a stack trace to standard error, why not throw a `SlimFixtureException` with message that date can not be parsed and `e` as cause?

src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
```diff
@@ -0,0 +1,64 @@
...
+        long timeStamp = 0;
+        SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
+
+        try {
+            timeStamp = sdf.parse(date).getTime();
+            if (!timestampHasMilliseconds) {
+                timeStamp = timeStamp / 1000L;
+            }
+        } catch (ParseException e) {
+            e.printStackTrace();
```
fhoeben: Instead of printing a stack trace to standard error, why not throw a `SlimFixtureException` with message that date can not be parsed and `e` as cause?

wiki/FitNesseRoot/HsacExamples/SlimTests/UtilityFixtures/DateFormatFixture.wiki
```diff
@@ -0,0 +1,36 @@
...
+!3 Examples
+
+!define dateTime {!today (dd-MM-yyyy HH:mm:ss)}
+!define date {!today (dd-MM-yyyy)}
+
+|script |date format fixture                             |
+|$stamp=|timestamp for date|${date}                      |
+|show   |format timestamp  |$stamp                       |
+|show   |format timestamp  |$stamp|as|yyyy/MM/dd HH:mm:ss|
+
```
fhoeben: Add a check that (with correct format) `$stamp` can be formatted to give `${date}`

wiki/FitNesseRoot/HsacExamples/SlimTests/UtilityFixtures/DateFormatFixture.wiki
```diff
@@ -0,0 +1,36 @@
...
+
+|script                    |date format fixture                                  |
+|timestamp has milliseconds|false                                                |
+|set date format           |dd-MM-yyyy HH:mm:ss                                  |
+|show                      |format date       |${dateTime}|as|yyyy/MM/dd HH:mm   |
+|$stamp=                   |timestamp for date|${dateTime}                       |
+|show                      |format timestamp  |$stamp                            |
+|show                      |format timestamp  |$stamp     |as|yyyy/MM/dd HH:mm:ss|
+
```
fhoeben: Add a check that (with correct format) `$stamp` can be formatted to give ${dateTime}

fhoeben:
Should there be a method to configure the timezone to use when converting timestamp to readable time?

src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Date;
+import java.util.TimeZone;
+
+/**
+ * Fixture that helps converting dates to and from various formats. Intended to be used as a utility fixture i.e. as library
+ */
+
+public class DateFormatFixture extends SlimFixture {
+    private String dateFormat = "dd-MM-yyyy";
+    private TimeZone timezone = TimeZone.getTimeZone("GMT+1");
```
fhoeben: Why not use `TimeZone.getDefault()`, might be more intuitive for people in different parts of the world?

src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
```diff
@@ -0,0 +1,73 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.TimeZone;
+
+/**
+ * Fixture that helps converting dates to and from various formats. Intended to be used as a utility fixture i.e. as library
+ */
+
+public class DateFormatFixture extends SlimFixture {
+    private String dateFormat = "dd-MM-yyyy";
```
fhoeben: Why not default to `((SimpleDateFormat) DateFormat.getDateInstance(DateFormat.SHORT)).toPattern();`, which uses the Java runtime default locale. For the Netherlands that would get you 'd-M-yy', which means no leading zeroes and year in two digits.
In tests you could/should than explicitly state the format to use, since that is probably not what you want exactly, but it is nicer for users in countries using a different format.

src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
```diff
@@ -0,0 +1,73 @@
...
+        } catch (ParseException e) {
+            throw new SlimFixtureException(false, "Could not parse " + date + " using format: " + dateFormat, e);
+        }
+        return formattedDate;
+    }
+
+    public long timestampForDate(String date) {
+        long timeStamp;
+        SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
+        sdf.setTimeZone(timezone);
```
fhoeben: Maybe a method to do the lines below would make the class a bit smaller?
```
        SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
        sdf.setTimeZone(timezone);
        return sdf;
```

src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
tcnh: I thought about using local defaults for timezone and format but decided to use defaults, beacouse using system settings yields unexpected results. For instance in one of my current projects, most testers create tests on their local machine, that has a dutch locale and a western europe timezone. The Jenkins master is a local server with a US locale and a western europe timezone and some jenkins slaves are EC2 instances with a US locale and a different timezone.

Setting a default ensures similar behaviour wherever the test runs. Your thoughts on that?

fhoeben: This is true. My first impulse is to explicitly set the desired values from the wiki in such a scenario. But in saying that the current solution works just as well, it also allows an explicit set for people that don't like the defaults coded in the class.

BUT, on the other hand, it is hard to back to using the system locale (as far as I know there is no string you can pass to explicitly use the system default, you have to use the constants from code).

So my preference is still to have the system defaults in code, and use these as defaults. This will be more intuitive to 'first-time' users. Where 'power-users' like yourself can set explicit values to ensure the test runs the same on any environment.

You could consider adding an extra constructor which explicitly configures the format and timezone, so you can still [use the fixture and configure it in a single table row](http://fitnesse.org/FitNesse.FullReferenceGuide.UserGuide.WritingAcceptanceTests.SliM.ConstructorArguments) (instead of needing 3).

What do you think?

fhoeben:
Did you notice https://github.com/fhoeben/hsac-fitnesse-fixtures/tree/enterDateField? I think that will greatly benefit from this fixture, so the user can ensure the date they want to enter is in yyyy-mm-dd format.
Do you have any comments/thoughts?

src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
tcnh: Agree. I like the extra constructor option so I'll implement that and set defaults to system default values.

wiki/FitNesseRoot/FrontPage/content.txt
```diff
@@ -13,3 +13,6 @@ To add your first "page", click the [[Edit][.FrontPage?edit]] button and add a [
 |[[Release Notes][.FitNesse.ReleaseNotes]]                     |''Find out about FitNesse's new features'' |
 
 !note Release ${FITNESSE_VERSION}
+
+
+${prop}
```
fhoeben: Why this change?

tcnh: Should not have been committed.

fhoeben:
I refactored a bit. Do you agree?

src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
```diff
@@ -0,0 +1,103 @@
...
+    protected Date parseDate(SimpleDateFormat sdf, String date) {
+        try {
+            return sdf.parse(date);
+        } catch (ParseException e) {
+            String pattern = sdf.toPattern();
+            throw new SlimFixtureException(false, "Could not parse " + date + " using format: " + pattern, e);
+        }
+    }
+
+    protected SimpleDateFormat getConfiguredDateFormat() {
```
tcnh: Maybe make this method public?
Use case: use the system default format in other tests as well (this may also help in converting for BrowserTest's new enterDate method?)

wiki/FitNesseRoot/HsacExamples/SlimTests/UtilityFixtures/DateFormatFixture.wiki
```diff
@@ -45,18 +45,33 @@ This fixture is intended to be used as a library. (See below)
...
+By default a library will use system default date format and timezone.
+
+|library            |
+|date format fixture|
+
+Example usage of date formatting methods from a string fixture script table:
+
+|script |string fixture                                  |
+|$stamp=|timestamp for date|${date}                      |
```
tcnh: If system defaults are used, we need to specify the input format if we want to ensure passing travis builds

fhoeben: You are soooo right (blush)...

src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
fhoeben: Sorry, I don't immediately grasp the use case yet. Why would one need to get the system default date format to get a value in yyyy-mm-dd?
One can just use a date format fixture instance to go from whatever format the date is in to yyyy-mm-dd, or ...?

wiki/FitNesseRoot/HsacExamples/SlimTests/UtilityFixtures/DateFormatFixture.wiki
fhoeben: That took me more effort than I expected..... ;-)

src/main/java/nl/hsac/fitnesse/fixture/slim/DateFormatFixture.java
tcnh: You're right. I was thinking of situations where one would get a date from a SUT, but the locale of the SUT has no relation to the locale of the machine running the test :)

Being able to use the date format fixture as a library should suffice for all date related use cases

PR #161, opened 2017-08-18
new selenium version 3.5.1

roydekleijn:


fhoeben:
Should we then not also update the IE driver (to 3.5.0, I don't believe 3.5.1 is available (via the maven plugin))?
I've created https://github.com/webdriverextensions/webdriverextensions-maven-plugin-repository/pull/14 to get 3.5.1 of the IE driver available

fhoeben:
IE driver 3.5.1 should be available now

PR #167, opened 2017-09-26
Added ExecuteJavascriptUntilIs(Not) methods, plus documentation and tests

tcnh:
Two methods to execute javascript until a value changes. (See it as an extension to the available executeJavascript method.)
This can be useful when refreshing data needs more than just a click or refresh.
For example when the application uses frames and refreshing the page would reset the frame that contains your data to a different state when refreshing. 

See if this is something worth making generally available. It does add a lot of versatility, but doesn't really honor the approach that one would want to script tests purely at the gui-level.


fhoeben:
I situations where I had to wait for Ajax calls that did not alter the DOM I found that executing some asynchronous Javascript (where Java would receive a callback when some Javascript event occurred) worked very well. This is the approach used in NgBrowserTest in 'waitForAngular', recently I implemented a similar approach for Richfaces.
That is maybe a bit more complex approach, but more elegant and efficient than polling.

Do you have a 'real world' example of the kind of Javascript you execute repeatedly?

But on the whole this looks good and I will merge it.

tcnh:
The acceptance test contains a real world example. I have a project where I have to refresh a single frame's (not iframe) content without refreshing the whole page, until a state changes. JavaScript is the only way I know how this can be achieved. The application has no functionality to trigger the refresh, other than going through a multi-click sequence.

PR #168, opened 2017-10-05
Added method to clear session storage

tcnh:
Shorthand to call sessionStorage.clear() javascript. (In line with clear local storage)
This has proven to be useful with an application that persisted data even after clearing cookies and local storage.


fhoeben:
The need to clear session storage surprises me. I believed session storage to be a subset of local storage. See for instance https://www.quora.com/What-is-the-difference-between-sessionstorage-localstorage-and-Cookies

I could image it being needed when you want to keep local storage but clear session storage...

tcnh:
They seem to be similar, but not the same. (At least in the app I was testing). 
See also: https://stackoverflow.com/questions/19867599/what-is-the-difference-between-localstorage-sessionstorage-session-and-cookies

PR #169, opened 2017-10-13
Added function to FileFixture to append a line to a file.

mourikwa:
Basically this pull request is about just one small utility function; "append a line to an existing file".
This exists in the slim filefixture class and the fileutil class.

I have also added a FileFixture junit test, for some functions, and one text file in resources which is being used in that test.

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/FileFixture.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/FileFixture.java
@@ -29,6 +29,10 @@ public class FileFixture extends SlimFixtureWithMap {
         }
     }
 
+    public String getDirectory(){
+        return directory;
+    }
+
     public String createContaining(String filename, String content) {
         String fullName = getFullName(filename);
         ensureParentExists(fullName);
@@ -194,5 +198,12 @@ public class FileFixture extends SlimFixtureWithMap {
 
     public boolean pollUntilSizeOfExceeds(String filename, long expectedSize) {
         return repeatUntil(new FunctionalCompletion(() -> exists(filename) && sizeOf(filename) > expectedSize));
+    }
+
+    public String appendContainingOnNewline(String filename, String content, boolean onNewLine) {
+        String fullName = getFullName(filename);
+        ensureParentExists(fullName);
+        File f = FileUtil.appendFile(fullName, content, onNewLine);
+        return linkToFile(f);
     }
 }
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/FileFixture.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/FileFixture.java
@@ -200,10 +200,17 @@ public class FileFixture extends SlimFixtureWithMap {
         return repeatUntil(new FunctionalCompletion(() -> exists(filename) && sizeOf(filename) > expectedSize));
     }
 
-    public String appendContainingOnNewline(String filename, String content, boolean onNewLine) {
+    public String appendTo(String extraContent, String filename) {
         String fullName = getFullName(filename);
         ensureParentExists(fullName);
-        File f = FileUtil.appendFile(fullName, content, onNewLine);
+        File f = FileUtil.appendToFile(fullName, extraContent, false);
+        return linkToFile(f);
+    }
+
+    public String appendToOnNewLine(String extraContent, String filename) {
+        String fullName = getFullName(filename);
+        ensureParentExists(fullName);
+        File f = FileUtil.appendToFile(fullName, extraContent, true);
         return linkToFile(f);
     }
 }
--- a/src/main/java/nl/hsac/fitnesse/fixture/util/FileUtil.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/util/FileUtil.java
@@ -26,7 +26,7 @@ public final class FileUtil {
      * @param filename file to read.
      * @return file's content.
      * @throws IllegalArgumentException if file could not be found.
-     * @throws IllegalStateException if file could not be read.
+     * @throws IllegalStateException    if file could not be read.
      */
     public static String loadFile(String filename) {
         ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
@@ -41,7 +41,7 @@ public final class FileUtil {
      * Copies UTF-8 input stream's content to a string (closes the stream).
      *
      * @param inputStream input stream (UTF-8) to read.
-     * @param fileName description for stream in error messages.
+     * @param fileName    description for stream in error messages.
      * @return content of stream
      * @throws RuntimeException if content could not be read.
      */
@@ -53,8 +53,8 @@ public final class FileUtil {
      * Copies first <numberOfLines> or all lines from UTF-8 input stream's content to a
      * string and closes the stream.
      *
-     * @param is input stream (UTF-8) to read.
-     * @param name description for stream in error messages.
+     * @param is            input stream (UTF-8) to read.
+     * @param name          description for stream in error messages.
      * @param numberOfLines number of lines to be read from input stream - if null, all lines are read
      * @return content of stream
      * @throws RuntimeException if content could not be read.
@@ -98,6 +98,7 @@ public final class FileUtil {
 
     /**
      * Copies source file to target.
+     *
      * @param source source file to copy.
      * @param target destination to copy to.
      * @return target as File.
@@ -125,7 +126,7 @@ public final class FileUtil {
      * Copy the contents of the given InputStream to the given OutputStream.
      * Closes both streams when done.
      *
-     * @param in the stream to copy from
+     * @param in  the stream to copy from
      * @param out the stream to copy to
      * @return the number of bytes copied
      * @throws java.io.IOException in case of I/O errors
@@ -159,7 +160,7 @@ public final class FileUtil {
      * Writes content to file, in UTF-8 encoding.
      *
      * @param filename file to create or overwrite.
-     * @param content content to write.
+     * @param content  content to write.
      * @return file reference to file.
      */
     public static File writeFile(String filename, String content) {
@@ -180,7 +181,7 @@ public final class FileUtil {
         return new File(filename);
     }
 
-    public static File  writeFromScanner(String filename, Scanner sc) {
+    public static File writeFromScanner(String filename, Scanner sc) {
         PrintWriter pw = null;
         try {
             pw = new PrintWriter(filename, FILE_ENCODING);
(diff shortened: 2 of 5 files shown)


fhoeben:
Did you notice the travis-ci run of the jUnit test failed?
Also, please run the code formatter you obviously used on the fixture code also on the unit test.

It would we nice if you could use the new method in the wiki sample/acceptance test of this fixture. This is where I (and many people) go for documentation what the fixture can do...

mourikwa:
> Did you notice the travis-ci run of the jUnit test failed?
Yes. I wasn't sure how to trigger a rerun. Used a new commit now.

> Also, please run the code formatter you obviously used on the fixture code also on the unit test.
Indeed; I didn't notice before committing that IDEA didn't format it automatically (this time). Done now.
 
> It would we nice if you could use the new method in the wiki sample/acceptance test of this fixture. This is where I (and many people) go for documentation what the fixture can do...
Agreed.  I will look into that.

src/main/java/nl/hsac/fitnesse/fixture/util/FileUtil.java
```diff
@@ -287,11 +289,41 @@ public static void ensureNoHtmlFiles(String dir) throws IOException {
...
     }
+
+    /**
+     * Appends content to file, in UTF-8 encoding.
+     * @param filename  file to create or overwrite.
+     * @param content   content to write.
+     * @param onNewLine whether a new line should be created before appending the content
+     * @return file reference to file.
+     */
+    public static File appendFile(String filename, String content, boolean onNewLine){
```
fhoeben: please name the method `appendToFile`

src/main/java/nl/hsac/fitnesse/fixture/util/FileUtil.java
```diff
@@ -287,11 +289,41 @@ public static void ensureNoHtmlFiles(String dir) throws IOException {
...
      * @param child file to get relative path for.
      * @return relative path.
      */
     public static String getRelativePath(String base, String child) {
         return new File(base).toURI().relativize(new File(child).toURI()).getPath();
     }
+
+    /**
+     * Appends content to file, in UTF-8 encoding.
+     * @param filename  file to create or overwrite.
```
fhoeben: not overwrite, but 'append to', right?

src/main/java/nl/hsac/fitnesse/fixture/util/FileUtil.java
```diff
@@ -287,11 +289,41 @@ public static void ensureNoHtmlFiles(String dir) throws IOException {
...
+     * @param content   content to write.
+     * @param onNewLine whether a new line should be created before appending the content
+     * @return file reference to file.
+     */
+    public static File appendFile(String filename, String content, boolean onNewLine){
+        PrintWriter pw = null;
+        try {
+            pw = new PrintWriter(new BufferedWriter(new FileWriter(filename, true)));
```
fhoeben: I don't believe this is UTF-8. Won't this get you the platform's file encoding?

src/test/java/nl/hsac/fitnesse/fixture/slim/FileFixtureTest.java
```diff
@@ -0,0 +1,129 @@
...
+    public void testTextIn() {
+        try {
+            fixture.setDirectory(testResourcesDir);
+            assertEquals("# Expected as first line of text.\n# Expected as 2nd line of text.",
+                    fixture.textIn(txtFilename));
+            fixture.textIn("foobar");
+        } catch (SlimFixtureException sfe) {
+            assertEquals("message:<<Unable to find: " + testResourcesDir + "foobar>>", sfe.getMessage());
+        } catch (IOException ioe) {
+            assertFalse("Should not happen.", true);
```
fhoeben: `fail()` is nicer than `assertFalse` with `true`

src/test/java/nl/hsac/fitnesse/fixture/slim/FileFixtureTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Test
+    public void testDeleteIfExists() {
+        try {
+            fixture.setDirectory(testResourcesDir);
+            if (fixture.exists(copyFilename)) {
+                fixture.deleteIfExists(copyFilename);
+            }
+        } catch (SlimFixtureException sfe) {
+            assertFalse("Should not happen: " + sfe.getMessage(), true);
```
fhoeben: see above

src/test/java/nl/hsac/fitnesse/fixture/slim/FileFixtureTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Test
+    public void testCopyTo() {
+        fixture.setDirectory(testResourcesDir);
+        assertTrue(fixture.exists(txtFilename));
+        try {
+            String res = fixture.copyTo(txtFilename, copyFilename);
+            assertEquals("<a href=\"file:" + testResourcesDir + copyFilename + "\">" + copyFilename + "</a>", res);
+        } catch (IOException ioe) {
+            assertFalse("Should not happen: " + ioe.getMessage(), true);
```
fhoeben: see above

src/test/java/nl/hsac/fitnesse/fixture/slim/FileFixtureTest.java
```diff
@@ -0,0 +1,129 @@
...
+            fixture.copyTo(txtFilename, copyFilename);
+            assertEquals("# Expected as first line of text.",
+                    fixture.takeFirstLineFrom(copyFilename));
+            assertEquals("# Expected as 2nd line of text.",
+                    fixture.takeFirstLineFrom(copyFilename));
+            fixture.textIn("foobar");
+        } catch (SlimFixtureException sfe) {
+            assertEquals("message:<<Unable to find: " + testResourcesDir + "foobar>>", sfe.getMessage());
+        } catch (IOException ioe) {
+            assertFalse("Should not happen: " + ioe.getMessage(), true);
```
fhoeben: see above

src/test/java/nl/hsac/fitnesse/fixture/slim/FileFixtureTest.java
```diff
@@ -0,0 +1,129 @@
...
+    public void testAppendContainingOnNewline() {
+        try {
+            fixture.setDirectory(testResourcesDir);
+            fixture.copyTo(txtFilename, copyFilename);
+            String res = fixture.appendContainingOnNewline(copyFilename, "Third line", true);
+            assertEquals("<a href=\"file:" + testResourcesDir + copyFilename + "\">" + copyFilename + "</a>", res);
+            assertEquals("# Expected as first line of text.\n# Expected as 2nd line of text.\nThird line",
+                    fixture.textIn(copyFilename));
+        } catch (IOException ioe) {
+            assertFalse("Should not happen: " + ioe.getMessage(), true);
```
fhoeben: see above

src/test/java/nl/hsac/fitnesse/fixture/slim/FileFixtureTest.java
```diff
@@ -0,0 +1,129 @@
...
+    public void testAppendContainingOnSameline() {
+        try {
+            fixture.setDirectory(testResourcesDir);
+            fixture.copyTo(txtFilename, copyFilename);
+            String res = fixture.appendContainingOnNewline(copyFilename, "Third line", false);
+            assertEquals("<a href=\"file:" + testResourcesDir + copyFilename + "\">" + copyFilename + "</a>", res);
+            assertEquals("# Expected as first line of text.\n# Expected as 2nd line of text.Third line",
+                    fixture.textIn(copyFilename));
+        } catch (IOException ioe) {
+            assertFalse("Should not happen: " + ioe.getMessage(), true);
```
fhoeben: see above

src/test/java/nl/hsac/fitnesse/fixture/slim/FileFixtureTest.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    @Test
+    public void testDelete() {
+        try {
+            fixture.setDirectory(testResourcesDir);
+            fixture.delete(copyFilename);
+        } catch (SlimFixtureException sfe) {
+            assertTrue("Should not happen: " + sfe.getMessage(), false);
```
fhoeben: see above

src/main/java/nl/hsac/fitnesse/fixture/slim/FileFixture.java
```diff
@@ -195,4 +199,11 @@ protected static FunctionalCompletion fileExistsCompletion(String fullname) {
     public boolean pollUntilSizeOfExceeds(String filename, long expectedSize) {
         return repeatUntil(new FunctionalCompletion(() -> exists(filename) && sizeOf(filename) > expectedSize));
     }
+
+    public String appendContainingOnNewline(String filename, String content, boolean onNewLine) {
```
fhoeben: I don't think this is the best method name for wiki usage. What do you expect to see in a wiki table row (that's also why extending the sample page will be nice).

I think something like `appendTo(String extraContent, String filename)` would be nicer, and maybe a separate `appendNewLineTo(filename)`, or `appendLineWithTo(content, filename)`. I think a separate method is nicer than always requiring a boolean parameter.

src/test/java/nl/hsac/fitnesse/fixture/slim/FileFixtureTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class FileFixtureTest {
```
fhoeben: Doesn't this unit test partially duplicate the content of 'FileFixture.wiki', which is executed as part of the acceptance test?
Do I overestimate the coverage of that test (aka no I'm wrong there is no duplication), this test adds ... as benefit, you missed that page, or ....?

mourikwa: As to whether the fixture tests are more useful when they exist as a Junit file or a FitNesse wiki table, I plead neutrality. I agree it may make more sense in this project to place the tests mostly or even only in FitNesse tables.
Is that what you mean and prefer?

fhoeben:
I do not really feel there should be NO unit test, but taking the trouble to write a new unit test for functionality already tested via the wiki is not something I would tend to do, or recommend. 
As I stated I like the tests via the wiki since the also serve as a 'more accessible' documentation of what the (utility) fixtures can do.

PR #175, opened 2017-11-24
extra test for the checkbsn functionality

maartenjanvangool:
made an extra unittest for the bsn functionality


src/test/java/nl/hsac/fitnesse/fixture/util/BsnUtilTest.java
```diff
@@ -2,24 +2,36 @@
 
 import org.junit.Test;
 
+import static junit.framework.TestCase.assertFalse;
```
fhoeben: `junit.framework` is deprecated. Maybe you can configure your IDE to only use the `org.junit...` versions?

maartenjanvangool: Was a mistake to import this one...Thanks for tip.

PR #186, opened 2018-01-04
Added methods to check if elements are not visible

tcnh:
Methods to check if elements are not visible, without having to wait for timeout on is visible. Handles possible async events as well.

Use case: we'd like to make sure something is not on the page.
We could use:

```| reject | is visible | something | ```

But that would cost [seconds before timeout] seconds to complete, which makes tests slow. Temporary setting the timeout to 1 second is a workaround, but still costs at least a full second, is not very readable and a bit error-prone. May also lead to timing issues in javascript intensive apps.

This PR adds methods to ensure something is not visible, using a waitdriver and the existing isVisibleImpl. Now we can use:

```| ensure | not visible | something | ```

Without worrying about timeouts.  This is also useful to dynamically wait for a spinner or toast message to disappear before continuing to the next instruction.

### CODE REVIEW

**Comment by tcnh** [comment]
Failing TravisCI check is addressed in https://github.com/fhoeben/hsac-fitnesse-fixtures/pull/187

File: src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java [hunk 1]
```diff
@@ -1539,6 +1539,52 @@ public boolean isVisibleOnPageIn(String place, String container) {
...
+     * @param container parent of place.
+     * @return true if the element is not on teh page (anymore)
+     */
+    @WaitUntil(TimeoutPolicy.RETURN_FALSE)
+    public boolean notVisibleOnPageIn(String place, String container) {
+        return notVisibleImpl(place, container, false);
+    }
+
+    protected boolean notVisibleImpl(String place, String container, boolean checkOnScreen) {
+        return waitDriver().until(d -> !isVisibleImpl(place, container, checkOnScreen));
```
fhoeben: Is the explicit use of `waitDriver()` required? Would the `@WaitUntil` not take care of repeating the wait?
Your new test passes if I change the line to `return !isVisibleImpl(place, container, checkOnScreen)`...

tcnh: Yes, you're right. Keep it simple :)
That also removes the need for the extra impl.

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests.wiki
```diff
@@ -54,14 +54,16 @@ The [[heuristic to find places][>FindingByHeuristics]] tries a number of transla
...
+|not visible <place> (in <container>)                                      |whether specified place is not visible in browser's window (used in combination with 'ensure' and 'reject'). Will retry for <seconds before timeout> seconds if place is visible.                                                                                                                                                                                                                                                                                                                                                                                                                      |
```
fhoeben: Can we merge these descriptions with 'is visible' and 'is visible on page'? It seems a bit too much to have 4 rows for these methods

tcnh: Merged. See if you think it's clearly dcumented like this

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests.wiki
```diff
@@ -52,16 +53,16 @@ The [[heuristic to find places][>FindingByHeuristics]] tries a number of transla
...
+|is/not visible <place> (in <container>)                                   |whether specified place is or is not visible in browser's window (used in combination with 'ensure'). Not visible will retry for <seconds before timeout> seconds if place is visible.                                                                                                                                                                                                                                                                                                                                                                                                                 |
```
fhoeben: Why not place 'not' in parentheses, like 'in <container>'?
I believe 'visible' will retry like 'Not visible', right?

PR #193, opened 2018-02-23
Updated Freemarker to latest version

tcnh:
Latest versions of freemarker allow for (among other things) auto-escaping of non-valid xml characters (such as ampersand) to reduce the risk of generating invalid requests.

See: https://freemarker.apache.org/docs/dgui_misc_autoescaping.html

We can now start a freemarker xml template with `<#ftl output_format="XML">` to enable automatic escaping.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,7 +97,7 @@
 		<dependency>
 			<groupId>org.freemarker</groupId>
 			<artifactId>freemarker</artifactId>
-			<version>2.3.23</version>
+			<version>2.3.27-incubating</version>
 		</dependency>
 		<dependency>
 			<groupId>org.apache.commons</groupId>


fhoeben:
`incubating` sounds like a non-production version. Should we really rely on that? Am I misinterpreting the meaning of the postfix?

tcnh:
In 2015, freemarker transferred to apache incubator and are using the postfix on all stable releases since then. 
From: https://freemarker.apache.org/freemarkerdownload.html

>Latest stable release: 2.3.27-incubating
>Released on 2017-11-03. Requires J2SE 1.5 or higher.
>
>This is a stable, final release. The "incubating" suffix is required by the Apache Software Foundation until the project becomes a fully accepted (graduated) Apache project. See disclaimer earlier and the Apache Incubator web site.

fhoeben:
Nice guys at Apache, why not update the groupId instead of the version if its not really 'apache' yet

PR #197, opened 2018-03-06
Added delete calls with a body.

keetron:


src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
```diff
@@ -191,16 +202,20 @@ public boolean postValuesTo(String serviceUrl) {
...
+
+    protected boolean sendToImpl(String body, String serviceUrl, String aContentType, String method) {
         boolean result;
         resetResponse();
         response.setRequest(body);
         String url = getUrl(serviceUrl);
         try {
-            storeLastCall("POST", serviceUrl);
+            storeLastCall(method, serviceUrl);
             getEnvironment().doHttpPost(url, response, headerValues, aContentType);
```
fhoeben: I think you will have to change the Environment and HttpClient also, right?
Currently we need to call different methods for different 'methods' and HttpClient does not yet support sending a body with a delete.

I know, there should be unit tests to check all this. Feel free to add...

src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
```diff
@@ -191,16 +202,20 @@ public boolean postValuesTo(String serviceUrl) {
...
-        return postToImpl(body, serviceUrl, getContentType());
+        return sendToImpl(body, serviceUrl, getContentType(), "POST");
     }
 
-    protected boolean postToImpl(String body, String serviceUrl, String aContentType) {
+    protected boolean deleteToImpl(String body, String serviceUrl) {
+        return sendToImpl(body, serviceUrl, getContentType(), "DELETE");
+    }
+
+    protected boolean sendToImpl(String body, String serviceUrl, String aContentType, String method) {
```
fhoeben: method as 1st param, please

PR #201, opened 2018-04-11
Added base64 encode/decode methods to string fixture to work with base64 (representation of) strings

tcnh:
2 convenience methods. Useful f.e. when you need to create auth headers at runtime in a test


fhoeben:
Are these better of in StringFixture, or should the corresponding methods in Base64Fixture be exposed?

tcnh:
I like them in String Fixture, but that's personal preference. I thought this might prevent confusion (I'd expect that I could send any Object to an encode method on Base64Fixture, instead of just Strings)

It's also more convenient in my practice, because I set StringFixture as a Library by default. I have no need for Base64Fixture in most projects.

fhoeben:
Currently Base64Fixture only supports encoding to/from files in the files section

fhoeben:
Base64 encoding of strings is not the most commonplace requirement I expect, in general that's more something I've encountered for binary data.
Can you elaborate on how this fits into a test, what is the en/decoded data used for? As input to other fixtures?

tcnh:
Mostly basic auth headers with http test. Alternatively we could add a method `set basic authorization header for user | x | and password | y |` in http test.

i.e. a credential comes from another service or from a system property on the build server. We need to translate these values to a base64 string to set as the Authorization header value in http test.

I have never needed to decode base64 strings in a test, but I thought it to be logical to support both ways and not just one.

tcnh:
Method is now in http test. Encode/decode removed from string fixture

fhoeben:
HsacAcceptanceTests.SlimTests.UtilityFixtures.StringFixtureTest now fails...

wiki/FitNesseRoot/HsacExamples/SlimTests/UtilityFixtures/StringFixtureTest.wiki
```diff
@@ -6,19 +6,19 @@ It is intended to be used is as a [[library][.FitNesse.UserGuide.WritingAcceptan
 
 This page demonstrates the features of the class (!-nl.hsac.fitnesse.fixture.slim.StringFixture-!) and therefore uses it on its own in a script table (instead of inside a script using a different fixture).
 
-|script|string fixture                                                                                    |
```
fhoeben: Can you revert this entire file to its previous, formatted, state?

src/main/java/nl/hsac/fitnesse/fixture/slim/StringFixture.java
```diff
@@ -3,6 +3,7 @@
 import nl.hsac.fitnesse.fixture.util.LineEndingHelper;
 import org.apache.commons.lang.StringUtils;
 
+import java.util.Base64;
```
fhoeben: Import can be removed

src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
```diff
@@ -749,6 +750,12 @@ public void setContentType(String aContentType) {
         contentType = aContentType;
     }
 
+    public void setBasicAuthorizationHeaderForUserAndPassword(String user, String password) {
+        String credential = user + ":" + password;
+        String base64credential = Base64.getEncoder().encodeToString(credential.getBytes());
```
fhoeben: Please add an explicit encoding (using 'ISO-8859-1') when converting the string to bytes.

PR #211, opened 2018-05-23
Fix the %d to resemble the actual bit prefix.

boredland:
When calling my tests binary, I get this stack
```
_EXCEPTION__:java.lang.IllegalStateException: The driver executable does not exist: /XXXX/fitnesse/chromedriver-linux-%dbit
```
I think there somehow was this digit-part in the code, the prevented fitnesse finding the actual executable `chromedriver-linux-64bit`.

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/ProjectDriverFactoryFactory.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/ProjectDriverFactoryFactory.java
@@ -65,9 +65,11 @@ public class ProjectDriverFactoryFactory {
     }
 
     protected String getExecutable(String basename) {
-        String name = getExecutableForOs(basename);
+        String name = null;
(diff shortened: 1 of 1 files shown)


boredland:
There actually is a /XXXX/fitnesse/chromedriver-linux-64bit.
When I add a symbolic link to /XXXX/fitnesse/chromedriver-linux-%dbit, it works. That made me think its a bug. When I use the fixture compiled from this PR, it works.

boredland:
I think this is the call for String.format(name, bit), that in my case doesn't replace the %d. What java version do you use? For me its OpenJdk 8.

fhoeben:
I believe I've used this with Oracle, IBM and OpenJDK. String.format with a %d is part of standard Java spec since Java 7.

fhoeben:
I still believe that the problem is that the executable is in the wrong the directory. Try placing it in a 'webdrivers' subdirectory (preferably below the wiki directory).

PR #213, opened 2018-05-24
Added geckodriver for linux

Romancha:


Romancha:
Thank you for totally explaining.
Now it doesn't matter.
Sorry for wasting your time.

PR #215, opened 2018-06-11
Date time calculator fixture

tcnh:
A simple (library) fixture to add/subtract years, months, days, hours, minutes, seconds to/from a given date.

Use case: we test some applications that set dates or times we'd like to validate in the system, relative to user input/events (f.e. end date automatically set x weeks from now, or interval timing based on event times)


maikelgithub:
there is already an option to do this

[check here for some examples](https://github.com/fhoeben/hsac-fitnesse-plugin/wiki/1.-Symbol-Samples)

fhoeben:
The currently existing features (only) work if you know beforehand/statically what the expected date is. This fixture allows you to calculate relative dates/times based on values obtained from system under test.

So they are similar, but not quite the same

PR #222, opened 2018-07-20
Ensure that a multimime response is handled correct

JaapD:
We had problems with a Soap with attachments response. This commit test is the response has a multiMime response works fine.

In the junit test is an example for such a response.

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/fit/SoapCallMapColumnFixture.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/fit/SoapCallMapColumnFixture.java
@@ -3,6 +3,14 @@ package nl.hsac.fitnesse.fixture.fit;
 import nl.hsac.fitnesse.fixture.Environment;
 import nl.hsac.fitnesse.fixture.util.XmlHttpResponse;
 
+import javax.mail.BodyPart;
+import javax.mail.MessagingException;
+import javax.mail.internet.MimeMultipart;
+import javax.mail.util.ByteArrayDataSource;
+import javax.mail.util.SharedByteArrayInputStream;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
 import java.util.Collections;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


fhoeben:
Thanks for the PR. A bit strange that mail dependencies are needed to process a SOAP response (I never encountered SOAP with attachments before).
One thing I do wonder about: would it be possible to integrate the updated response handling in XmlHttpResponse, or at least provide methods to extract parts there? Now you addressed the problem only for people using fit to test SOAP services, people using slim (via XmlHttpTest) are not able to benefit from your fix.

src/main/java/nl/hsac/fitnesse/fixture/fit/SoapCallMapColumnFixture.java
```diff
@@ -131,4 +141,32 @@ protected String xmlAsHtml(String xml) {
...
+                response.setResponse(xml);
+            }
+        } catch (MessagingException | IOException e) {
+            throw new RuntimeException("unparseble response", e);
+        }
+    }
+
+    private String getBodypartString(BodyPart bp) throws IOException, MessagingException {
+        SharedByteArrayInputStream stream = (SharedByteArrayInputStream) bp.getContent();
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(stream))) {
```
fhoeben: Isn't there a convenient method in the apache libraries (e.g. IOUtils) to copy a stream to a string?

JaapD:
It is a good idea to modify XmlHttpResponse. What would be the best place, override setResponse()?

fhoeben:
I think a correct approach will allow users to choose whether they want to get a certain body part as response (so not always 0 like currently, and not always but only when requested), and ideally also allow to check how many parts are available.

So maybe add a new method like `boolean selectBodyPartAsResponse(int index)`, which will set the response to the correct part. The full multipart message should also saved somewhere so that a test could retrieve multiple parts (e.g. first `0` then `1`, etc). And maybe index `-1` could reset the response to the full response.
Additionally `int bodyPartCount()` could return the number of parts.

You could then have SoapCallMapColumnFixture automatically select the first element if there are more than 0, like you do in the current code.

This way the feature is available, but I believe it will be usable in more situations where the response is a bit different. It is the fixture that decides to trigger the selection of the first element, but the response class contains the implementation on how to work with the multipart.

JaapD:
I changed the XmlHttpTest, not sure where to modify the SoapCallMapColumnFixture.
The only MimeMultipart I found was in the mail dependencies, that is why I used them.

PR #230, opened 2018-08-14
RichFaces ajax can be inside the eventhandler instead of starting it,…

karsie:
… for example starting with jsf.util.chain

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/web/RichFacesTest.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/web/RichFacesTest.java
@@ -156,8 +156,8 @@ public class RichFacesTest extends BrowserTest<WebElement> {
     protected boolean eventTriggersAjax(WebElement element, String attribute) {
         String eventHandler = element.getAttribute(attribute);
         return eventHandler != null
-                && ajaxStartWithOnly? eventHandler.startsWith(RICH_FACES_AJAX_CALL)
-                                    : eventHandler.contains(RICH_FACES_AJAX_CALL);
+                && (ajaxStartWithOnly? eventHandler.startsWith(RICH_FACES_AJAX_CALL)
+                                    : eventHandler.contains(RICH_FACES_AJAX_CALL));
     }
 
     public void waitForJsfAjax() {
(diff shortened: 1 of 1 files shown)


src/main/java/nl/hsac/fitnesse/fixture/slim/web/RichFacesTest.java
```diff
@@ -160,10 +160,10 @@ protected void waitForJsfAjaxImpl(boolean checkLocation) {
         try {
             // if jsf is present on page, add an event listener that will be triggered when next Ajax request completes
             if (checkLocation) {
-                waitForJavascriptCallback("if(!window.jsf||window.location.href!==arguments[0]){callback();}else{jsf.ajax.addOnEvent(function(data){if(data.status!='begin')callback();});}",
+                waitForJavascriptCallback("if(!window.jsf||window.location.href!==arguments[0]){callback();}else{jsf.ajax.addOnEvent(function(data){if(data.status=='success')callback();});}",
                         previousLocation);
             } else {
-                waitForJavascriptCallback("if(!window.jsf){callback();}else{jsf.ajax.addOnEvent(function(data){if(data.status!='begin')callback();});}");
+                waitForJavascriptCallback("if(!window.jsf){callback();}else{jsf.ajax.addOnEvent(function(data){if(data.status=='success')callback();});}");
```
fhoeben: I explicitly waited on anything else than status 'begin' to ensure we also stop waiting when an error is returned by the ajax call. Does the current change still stop waiting on error (what other statuses exist than success and begin)?

fhoeben: Agreed

src/main/java/nl/hsac/fitnesse/fixture/slim/web/RichFacesTest.java
```diff
@@ -145,7 +145,7 @@ protected void storeLocationBeforeAction() {
 
     protected boolean eventTriggersAjax(WebElement element, String attribute) {
         String eventHandler = element.getAttribute(attribute);
-        return eventHandler != null && eventHandler.startsWith("RichFaces.ajax(");
+        return eventHandler != null && eventHandler.contains("RichFaces.ajax(");
```
fhoeben: I did not do this before because if the attribute does not start with the call we cannot be sure that the Ajax call will actually be executed (the might be an `if` before it causing no Ajax call to be sent). And in that case waiting for completion of a call that is not sent will not end so it will break the test.

Is that something that cannot happen, so an unfounded fear?

PR #247, opened 2019-01-25
Select As clears multiple selectboxes first

karsie:
selectAs clears multiple selectboxes first
Added selectAsIn, similar to selectForIn


fhoeben:
Looks quite good.
Could you add an acceptance test, and list the command in the BrowserTests documentation page (below HsacExamples)?

karsie:
How come I only see this error when running in the AcceptanceTests Suite? Running it solo works perfectly

fhoeben:
It looks like you have a problem with the 'list rendering' in the wiki (they are being displayed with numbers instead of comma separated).

The alternative list rendering can be left over from another acceptance test. Side effects are not very nice.

Can you try adding the following to the top of your test page?
```
|script|list fixture|
|display lists standard|
```

karsie:
Thanks, Updated

PR #248, opened 2019-01-25
Use new ChromeDriver(ChromeOptions) constructor

karsie:


fhoeben:
Doesn't seem to work in the Travis-CI build. Is this some kind of flaky test result, does it work locally (also when running from jUnit)

karsie:
Seems to fail on the part where acceptance-tests passes headless to chrome, which is different when using ChromeOptions directly. I don't see a normal way of getting it to work unless refactoring the profile part in a bigger way, so I'll drop it.

PR #249, opened 2019-02-04
Added xsd schema validation to XMLValidator and XmlFixture

tcnh:
This PR enables XmlFixture to validate xml against a schema (xsd) from a file.
The logic is in XmlValidator, but in all use cases I've seen it makes sense to call it from XmlFixture, so I added a fixture method to it that uses the method in XmlValidator.

Commit also contains unit tests, an example in the wiki documentation and files necessary for testing.


src/test/resources/temp-copy.txt
```diff
@@ -0,0 +1,2 @@
+# Expected as first line of text.
```
fhoeben: Why is this file present?

src/main/java/nl/hsac/fitnesse/fixture/util/XMLValidator.java
```diff
@@ -53,4 +65,30 @@ public void isWelformed(String xml) {
...
+
+    /**
+     * Validate currently loaded xml against an xsd
+     * @param xsdFileName the XSD schema file to use for validation
+     * @return true if the xml validates against the schema.
+     * Throws a descriptive SlimFixtureException when validation fails
+     */
+    public boolean validateAgainst(String xmlContent, String xsdFileName) {
+        try {
+            String xsdContent = new FileFixture().textIn(xsdFileName);
```
fhoeben: I try not to uses fixtures from util classes. Can you move the reading of the file to XmlFixture, and just supply the xsdContent as parameter here.
You could have an overload that reads a file, but that can only use FileUtil.

src/main/java/nl/hsac/fitnesse/fixture/slim/XmlFixture.java
```diff
@@ -90,6 +90,15 @@ public String createFileFromBase64ContentOf(String baseName, String xPath) {
         }
     }
 
+    /**
+     * Validat the loaded xml against a schema in file xsdFileName
+     * @param xsdFileName filename of the xsd to use
+     * @return true if the xml validates against the schema. Throws a descriptive exception otherwise
+     */
+    public boolean validateAgainst(String xsdFileName) {
+        return new XMLValidator().validateAgainst(content, xsdFileName);
```
fhoeben: Please extract the content from the xsd using FileFixture here, and not in the util class.

src/main/java/nl/hsac/fitnesse/fixture/slim/XmlFixture.java
```diff
@@ -90,6 +90,15 @@ public String createFileFromBase64ContentOf(String baseName, String xPath) {
         }
     }
 
+    /**
+     * Validat the loaded xml against a schema in file xsdFileName
+     * @param xsdFileName filename of the xsd to use
+     * @return true if the xml validates against the schema. Throws a descriptive exception otherwise
+     */
+    public boolean validateAgainst(String xsdFileName) {
```
fhoeben: Does it also make sense to provide a method where the xsd content is provided by the wiki (preformatted)?

src/test/resources/temp-copy.txt
tcnh: Good question! Removed

src/main/java/nl/hsac/fitnesse/fixture/util/XMLValidator.java
tcnh: Moved file reading to xmlFixture/xmlHttpTest classes

wiki/FitNesseRoot/HsacExamples/SlimTests/HttpTests/XmlHttpTestResponseSchemaValidation.wiki
```diff
@@ -0,0 +1,54 @@
+---
+Test
```
fhoeben: This test seems more appropriate for the HsacAcceptanceTests, so we run it always.

A single example (with an external XML file?) is also nice, but ensuring it works for every build....

wiki/FitNesseRoot/HsacExamples/SlimTests/HttpTests/XmlHttpTestResponseSchemaValidation.wiki
```diff
@@ -0,0 +1,54 @@
...
+---
+|script           |mock xml server setup                             |
+|add response file|http://files/xmlTestExamples/valid_shiporder.xml  |
+|add response file|http://files/xmlTestExamples/invalid_shiporder.xml|
+|$url=            |get mock server url                               |
+
+!*> XSD String definition
+!define xsdSchema { {{{<?xml version="1.0" encoding="UTF-8" ?>
```
fhoeben: The test might be more maintainable if you use FileFixture in wiki page explicitly to load the file into a Slim symbol, and pass that to the xml http fixture method,  instead of repeating the xsd contents here.

wiki/FitNesseRoot/HsacExamples/SlimTests/HttpTests.wiki
```diff
@@ -17,6 +17,7 @@ The >HttpPost2UsingScenarioTest makes multiple SOAP calls to a service, where sc
 >HttpHeadTest shows an example HEAD request, retrieving only response status and headers.
 >HttpCompressionTest shows an example GET request with gzip compression support.
 >HttpsWithoutCertificateVerification shows example GET requests without SSL certificate verification
+>XmlHttpTestResponseSchemaValidation shows an example of how we can validate response xml agains a given XSD schema
```
fhoeben: This page no longer exists in 'examples', so this line should be removed also

PR #253, opened 2019-02-27
HTML5 compatible drag and drop method

tcnh:
Implemented workaround for https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/3604 to be able to use html5 drag and drop with browser test using Javascript from https://github.com/Photonios/JS-DragAndDrop-Simulator

AccepatnceTest added, BrowserTest documentation updated


fhoeben:
@roydekleijn this seems to address the same problem you contacted me about some time ago. Does this fix also work in your context?

fhoeben:
Now there is a separate method to do html5 drag and drop. Would it not be possible to automatically detect whether we need 'normal' or 'html5' drag and drop? For instance by checking whether the element to be dragged has attribute `draggable` with value true?

tcnh:
I like the idea to auto-detect, but I also see some potential hazards. HTML5 drag & drop is not limited to the draggable=true attribute, but in most browsers, text selections, img elements, and anchor elements with an href attribute are draggable by default. This behaviour can however easily be overidden using all kinds of javascript frameworks.

We can try doing the auto-detect but there may be lots of corner cases that could potentially break it.

fhoeben:
Would it make sense to keep the html5 version as separate method, but route the original/existing method to the html5 version if 'draggable=true'?

What happens if you use the html5 version on other drag and drop (or the other way around) nothing or an error?

tcnh:
We'd have to test, but I expect no error, just nothing happening

tcnh:
Updated to reflect the above.
Using html5 method on non-html5 elements and v.v. behaves like before: no error, no change on page.

updated acceptance test with automatic html5 switch example and updated browsertest documentation to reflect this behaviour.

roydekleijn:
> @roydekleijn this seems to address the same problem you contacted me about some time ago. Does this fix also work in your context?

Thanks for the mention! This fixed our HTML5 issue. We started implementing something similar, but that's nolonger needed. Thanks again.

PR #254, opened 2019-03-08
list of one returns value

maartenjanvangool:
when a jsonpath returns a list of one, instead of an exception, the 1 value is returned and can be evaluated.

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/util/JsonPathHelper.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/util/JsonPathHelper.java
@@ -12,13 +12,14 @@ import java.util.List;
 
 /**
  * Helper to evaluate JsonPath expressions against a JSON object.
+ *
  * @link https://github.com/jayway/JsonPath
  * @link http://jsonpath.herokuapp.com/
  */
 public class JsonPathHelper {
     private final static Configuration CONF = Configuration
-                                                .defaultConfiguration()
-                                                .addOptions(Option.SUPPRESS_EXCEPTIONS);
+            .defaultConfiguration()
+            .addOptions(Option.SUPPRESS_EXCEPTIONS);
     private final static ParseContext CONTEXT = JsonPath.using(CONF);
 
     private String lastJson;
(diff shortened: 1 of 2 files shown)


fhoeben:
There already was a method to extract the n-th element if there are multiple results. That could already be used to retrieve the first value. 
What is the added value of (also) extracting the first value implicitly?

tcnh:
It's inconvenient (and confusing) that json path returns a list of one item in many cases when using dot notation, instead of just the wanted key's value.

For example when querying the json in the unit test in this PR (see https://jsonpath.herokuapp.com/) Using path `$..book[2].author` returns a list of one String (the vaue of the key author), resulting in the exception that this query returns multiple items ant not one. Which is not the case, because there is exactly one result.

This can be circumvented by using only normalize expresiions like `$['store']['book'][2]['author']`, but that is not possible when doing conditional paths like `$..book[?(@.title == 'Sword of Honour')].author`

fhoeben:
I was thinking more that one could use the command `element <index> of json path <expression>`.

So for the example you give one could already do:
`|check|element |0| of json path |$..book[2].author|Herman Melville|`

maartenjanvangool:
There are ways to get what you want, but the exception message I got with the current implementation is not correct. I did not get multiple items, I got a list of one item. It seems to me either the functionality is changed according to what I suggested, or the exception message is changed. 

I would prefer the first option; convenience by itself seems to me a pretty good reason to add something. Having to extract the first item of a list of one when your json path can only return one item seems to me a bother not worth having. Especially combined with a confusing exception.

fhoeben:
My problem is that the expression you use may return multiple results (according to the json path parser), whether or not it actually returns a single element depends on the document document.
I don't really like to change the behaviour based on whether a single result or multiple are returned.

So you are right the exception message is not correct: it should state that multiple results MIGHT be returned by the expression not that multiple ARE returned. 

If you want to (or must) use an expression that may return multiple results I believe it is better to make the test case express that you expect only one result, and what that first result is. (Your statement 'when your json path can only return one item ' is incorrect: it can return multiple you expect it to return only one). 
It would be great if json path allowed you to express that you want to first result of a conditional path in the expression. But I believe that is not possible.

So I don't think it's the fixtures job to change its behaviour based on the number of results of a json path expression it should always act the same, evaluating a json path according to the spec.

maartenjanvangool:
I'll make a new pull request to change the exception message...

violinner:
Using jsonHttpTest, this issue makes the following sequence unworkable:

    |script | json http test | 
    |post | my_query | to | $my_url |
    |repeat until json path | $.items[?(@.a=="v1" && @.b=="v2")].ccc | is | 42 |

fhoeben:
@violinner I'm not a 100% sure, but shouldn't the expected path be a list containing a single element: 42? So something along the lines of 
```
|repeat until json path | $.items[?(@.a=="v1" && @.b=="v2")].ccc | is | [42] |
```

I haven't checked whether that works, but that is the kind of check I would expect one should be able to write

tcnh:
That will not work, as th repeat until function also relies on the jsonPath function. This will always throw an exception if a list is returned (JsonPathHelper:36). Json paths containing a search expression `?(@.xyz)` always return a list. Fixing that at the cost of some consistency was the point of this original PR

fhoeben:
I saw this PR originally as a way to automatically return the first element if the expression is not definite. I'm not in favour of that. But thinking about it now I believe automatically returning a list with all results in that case would be good.
For normal checks this offers only a small advantage, no need to use special method for multiple results, but for the `repeat until` it also removes the need to create special methods.

PR #269, opened 2019-06-19
Json javadoc for contexthelp

tcnh:
Have javadoc json available in the jar's meta-inf, so autocompleteresponder can serve all relevant info to bootstrap-plus.

See https://gitlab.com/praegus/toolchain-fixtures/fixture-doc-generator-doclet for source and readme

To test:
Add the following execution to your build's maven dependency plugin:

```
 <execution>
	<id>copy-plugins</id>
	<phase>generate-resources</phase>
	<goals>
		<goal>copy</goal>
	</goals>
	<configuration>
		<artifactItems>
			<artifactItem>
				<groupId>nl.praegus</groupId>
				<artifactId>toolchain-fitnesse-plugin</artifactId>
				<version>1.15</version>
				<classifier>jar-with-dependencies</classifier>
				<overWrite>true</overWrite>
			</artifactItem>
		</artifactItems>
		<outputDirectory>${project.basedir}/wiki/plugins</outputDirectory>
	</configuration>
</execution>
```
To be able to use the plugins' 'debug script' table, make sure that `nl.praegus:toolchain-fixtures:1.14` is added as a dependency, `nl.praegus.fitnesse.slim.fixtures.util` is imported in slim and `pause test fixture` is added as a library.

Don't forget  to register bootstrap-plus as the theme (this will no longer be necessary from the next fitnesse and toolchain-plugin releases)

Now when editing a test, your context helper (top right) will contain all javadoc as documentation, as well as the names of all expected parameters, public constructors. These will also be available in codeMirror's autoComplete list and bootstrap-plus' validator.

Special pages or new test pages will get the context of the last test that was opened.


fhoeben:
The new version of FitNesse has been released. Do you need to update the plugin to take advantage of that?

Does the toolchain-fitnesse-plugin-with-dependencies also include the toolchain-fixtures? Or do we need to add that separately as a dependency (besides the copy action you specified)?

pom.xml
```diff
@@ -504,6 +504,42 @@
...
+				</executions>
+
+				<configuration>
+					<doclet>nl.praegus.doclets.FixtureDocGenerator</doclet>
+					<docletArtifact>
+						<groupId>nl.praegus</groupId>
+						<artifactId>fixture-doc-generator</artifactId>
+						<version>1.0</version>
+					</docletArtifact>
+					<excludePackageNames>nl.hsac.fitnesse.sample:
```
fhoeben: Does the plugin also support an 'include' configuration? Just specifying what to include seems more compact, and less error-prone, that listing everything to exclude.

tcnh: There is a SourceFileIncludes option: https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#sourceFileIncludes

Would have to look into that, but I think we can specify patterns of classes (but not packages) to include.

tcnh:
>The new version of FitNesse has been released. Do you need to update the plugin to take advantage of that?

Yes. I already have a branch with the necessary changes. I'll merge that later today.

>Does the toolchain-fitnesse-plugin-with-dependencies also include the toolchain-fixtures? Or do we need to add that separately as a dependency (besides the copy action you specified)?

Yes, it nedds to be added as a dependency. Including it (or just pauseTestFixture) in the plugin jar would mean we have to add the plugin Jar to the slim classpath, which might not be very clear when configuring your project.

pom.xml
fhoeben: Being able to specify patterns of classes to include (similar to the current `excludePackageNames`) would be my preference.

pom.xml
```diff
@@ -504,6 +509,42 @@
...
+				<executions>
+					<execution>
+						<phase>generate-resources</phase>
+						<goals>
+							<goal>javadoc</goal>
+						</goals>
+					</execution>
+				</executions>
+
+				<configuration>
```
fhoeben: Can this configuration be included inside the `execution` tag above to ensure it is not applied in the execution defined in the release profile?

PR #274, opened 2019-08-21
Make scroll to element behaviour configurable

tcnh:
This PR adds an option to scroll elements to center of viewport. Defult behaviour is as it currently is.

ratio: we regularly encounter cases where elements are unclickable, even after a scroll into view, because menubars floating on top in a fixed position obscure the element after scrolling. 
By setting the new option using `| scroll elements to center |`, the scrollTo method in seleniumHelper will scroll the element to the center of the viewport, where it's less likely to encounter menu-bars or footers.

The js configuration used is a scrollIntoViewOptions object, as described on https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView 

Includes an acceptance test/example and updated BrowserTest documentation.


fhoeben:
I see that on Travis one of the tests is failing. Running the test locally by itself it seems fine. Any thoughts to why that would be? All I can think of is maybe a viewport size issue (i.e. do we need to set window size)?

I like the feature, I faced the problem myself and worked around it by choosing different scroll targets. But I feel the actual methods need to change a bit. Instead of the method to set it to center and another to set it to default the same pattern should be followed as used in for instance `trimOnNormalize` and `setTrimOnNormalize`. By having a 'get current' and 'set' method you can have tests that set the behaviour to what they need and then switch it back to what it was. Your current implementation does not support that as there is no way to get the current value of the setting.

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
```diff
@@ -961,9 +961,11 @@ public T findElement(SearchContext context, By by) {
     /**
      * Trigger scrolling of window to ensure element is in visible.
      * @param element element to scroll to.
+     * @param toCenter boolean value indicating wether to center the element in the viewport or use the automatic setting
      */
-    public void scrollTo(WebElement element) {
-        executeJavascript("arguments[0].scrollIntoView(true);", element);
+    public void scrollTo(WebElement element, boolean toCenter) {
+        String scrollIntoViewArgs = toCenter ? "{behavior: \"auto\", block: \"center\", inline: \"center\"}" : "true";
+                executeJavascript("arguments[0].scrollIntoView("+ scrollIntoViewArgs +");", element);
```
fhoeben:
```suggestion
                executeJavascript("arguments[0].scrollIntoView(" + scrollIntoViewArgs + ");", element);
```

tcnh:
Agree. Changed the implementation to get/set.

The failing test seems to always fail when run in the compete suite.. When run alone it succeeds. I'll look into that after lunch.  It -seems- like the y-position on the page from the previous test is kept, effectively rendering the first few lines outside the viewport.

PR #279, opened 2019-09-17
Clarification webdriver incompatibility

maartenjanvangool:
Current message when driver is not compatible:
__EXCEPTION__:ABORT_SLIM_TEST:nl.hsac.fitnesse.fixture.slim.StopTestException: Unable to create selenium session using: nl.hsac.fitnesse.fixture.util.selenium.driverfactory.LocalDriverFactory@7ca33c24

After this pull request is accepted, the message is something like:
__EXCEPTION__:ABORT_SLIM_TEST:nl.hsac.fitnesse.fixture.slim.StopTestException: session not created: This version of ChromeDriver only supports Chrome version 74
  (Driver info: chromedriver=74.0.3729.6

Seems a lot better to me.

I also added some unittests and a mocking framework in order to be able to test driver manager


src/test/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverManagerTest.java
```diff
@@ -0,0 +1,44 @@
...
+    public void getSeleniumHelperReturnsSeleniumHelper() {
+        when(driverFactory.createDriver()).thenReturn(webDriver);
+
+        SeleniumHelper result = driverManager.getSeleniumHelper();
+
+        Assert.assertNotNull(result);
+        Assert.assertEquals(result.driver(), webDriver);
+    }
+
+    @Test(expected = StopTestException.class)
```
fhoeben: Given that you want a better message, shouldn't you test the message in the exception?

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverManager.java
```diff
@@ -69,7 +73,7 @@ public void setSeleniumHelper(SeleniumHelper helper) {
         this.helper = helper;
     }
 
-    protected SeleniumHelper createHelper(WebDriver driver) {
+    protected SeleniumHelper createHelper() {
```
fhoeben: The driver was passed to this method so it could be used by subclasses (specifically the `nl.hsac.fitnesse.fixture.util.mobile.AppiumDriverManager` uses it to determine what type of device is used in tests. Please leave it in place.

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverManager.java
```diff
@@ -24,6 +24,10 @@ public DriverManager() {
         Runtime.getRuntime().addShutdownHook(new Thread(() -> closeDriver()));
     }
 
+    protected DriverManager(DriverFactory driverFactory) {
```
fhoeben: Why is the extra constructor needed? I believe Mockito will also inject mocks directly into (private) fields, so it does not need it

src/test/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverManagerTest.java
maartenjanvangool: looked up how to do that with junit, added. Have you ever considered using assertj? It's a much better library for assertions, and has a nice one for checking exceptions as well. I find the junit assertions a bit verbose.

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverManager.java
maartenjanvangool: Put it back in. I will say this implementation has a smell; but since it's irrelevant for this pr, let's leave it at that.

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverManager.java
maartenjanvangool: Apparently :-).

src/test/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverManagerTest.java
fhoeben: I've used assertj in the past and its assertions, especially for exceptions are much better that the standard jUnit stuff.  I tend not to use the jUnit `@Rule` stuff and just do a `try/catch` in the test method when just using jUnit.
Just like I didn't add mockito before: I never felt the complexity of the test warrant adding assertj, since I also did not feel like re-writing the existing tests. PRs to clean stuff up are always welcome.... ;-)

maartenjanvangool:
Thanks for merging!

PR #280, opened 2019-10-24
Better httptest responses

maartenjanvangool:
I'm not entirely sure whether or not this is the best solution, but I found it very confusing to have a red "get from" without knowing what the issue is in http test. Now you get an actual message: 

image

I fixed the unittests with regard to repeating, and the repeating itself I think. I thought it convenient to now add assertj, since I had to assert an exception. Tell me if you like that. 

Perhaps this one needs some more, but @fhoeben, what do you think?


fhoeben:
I haven't looked at your changes in detail, but can one still test for not ok status codes? Previously if you wanted to test that a certain call responded with a 404 you could do:
```
|reject|get from| ...|
|check|response code|404|
```
That would be 'green'
Can this still be done, or is the exception always reported?

maartenjanvangool:
I saw i wasnt' working with the latest version, so I made some more changes. I think we should try to get rid of the 'validResponse' method above, throwing a runtime exception is quite the smell imo. But I dared not touching that function yet.

maartenjanvangool:
> I haven't looked at your changes in detail, but can one still test for not ok status codes? Previously if you wanted to test that a certain call responded with a 404 you could do:
> 
> ```
> |reject|get from| ...|
> |check|response code|404|
> ```
> 
> That would be 'green'
> Can this still be done, or is the exception always reported?

Hmm no since an exception is thrown this won't work. I guess this makes this a breaking change, which is not desired... Is there a way to report the error while still being able to do this?

fhoeben:
You could make some sort of 'toggle' what to do when an error is returned (by default return false, if toggle enabled: throw exception).
I recently did something similar to prevent test being aborted if no connection to server could be made...

maartenjanvangool:
Ok, build is green now. I have put the default value of the toggle to false. In this way the change would be backward compatible. I would prefer to set the default to true; testing a non-valid response is not the default I would say; but again, this would have backward compatibility issues. 

@fhoeben what do you think?

fhoeben:
The thinking for returning false was that it highlights the response indicates the request was not a success, but sometimes in testing that is exactly what one wants to verify. 
Checks after the call (e.g. on the response status or body) can provide any extra needed information on why the call was not a success. 

But then again sometimes one just wants to fire the call and not bother with extra check to get the details if it fails, in that case the exception is easier as it allows you to supply the extra information without the need to clutter your tests with assertions that you expect you never need.

So I think the toggle approach is the way to go, and default false provides backwards compatibility. So that all sounds pretty good. I hope to be able to dive in the code now.

P.S. I really like assertj but introducing it is probably best in a separate PR

src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
```diff
@@ -24,7 +24,9 @@
  * Fixture to make HTTP requests using Slim scripts and/or scenarios.
  */
 public class HttpTest extends SlimFixtureWithMap {
-    /** Default content type for posts and puts. */
+    /**
```
fhoeben: reformatting JavaDoc seems to make this PR a bit hard to read.
I'm not really bothered about fixing these, but if you are please make a PR for the entire project.

src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
```diff
@@ -872,6 +901,7 @@ public boolean stopTestOnException() {
 
     // Polling
     public boolean repeatUntilResponseStatusIs(final int expectedStatus) {
+        doNotThrowExceptionOnHttpRequestFailure();
```
fhoeben: Changing the value of the setting seems like an unexpected side effect of repeatUntil.
Please ensure the toggle is set back to its previous setting after the 'repeat'

I actually believe a user that would want to use the repeat feature probably set the 'do not throw' setting before the first call (before the initial call that is repeated by calling one of the 'repeat' methods). So toggling here in code can probably be removed altogether.

src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
```diff
@@ -37,9 +39,19 @@
     private String contentType = DEFAULT_POST_CONTENT_TYPE;
     private String lastUrl = null;
     private String lastMethod = null;
+    private boolean throwExceptionOnHttpRequestFailure = false;
+
+    public void doThrowExceptionOnHttpRequestFailure() {
```
fhoeben: In general I prefer to offer a method to get the current value and one to change it. (this also allows for usage in scenarios where one first saves the current value in a symbol, changes the setting to the desired state and then resets it to the initial state, without a need to know the initial state)

src/main/java/nl/hsac/fitnesse/fixture/slim/SlimFixture.java
```diff
@@ -168,8 +168,9 @@ protected boolean repeatUntil(RepeatCompletion repeat) {
                 waitMilliseconds(nextInterval);
 
                 loopTimer.start();
-                repeat.repeat();
+
                 try {
+                    repeat.repeat();
```
fhoeben: Doesn't moving this line break backwards compatibility in situations where the 'repeat' throws an exception for another reason (e.g. unable to connect to host)

src/main/java/nl/hsac/fitnesse/fixture/util/HttpResponse.java
```diff
@@ -31,12 +32,35 @@ public void validResponse() {
...
+            throw new RuntimeException("Server reported client error: " + statusCode);
+        }
         if (statusCode >= 500 && statusCode <= 599) {
             throw new RuntimeException("Server error returned: " + statusCode);
         }
-        if (statusCode >= 400 && statusCode <= 499) {
-            throw new RuntimeException("Server reported client error: " + statusCode);
+    }
+
+    public boolean responseIsValid(boolean throwExceptionOnHttpRequestFailure) {
```
fhoeben: why not just use the implementation of the existing method and catch the RuntimeException and convert that to a SlimFixtureException if boolean is true

src/main/java/nl/hsac/fitnesse/fixture/util/XmlHttpResponse.java
```diff
@@ -22,7 +23,25 @@
...
+    }
+
+    @Override
+    public boolean responseIsValid(boolean throwExceptionOnHttpRequestFailure) {
+        boolean responseIsValid = super.responseIsValid(throwExceptionOnHttpRequestFailure);
+        if (throwExceptionOnHttpRequestFailure) {
+            checkXmlResponse();
+        } else {
+            try {
+                checkXmlResponse();
```
fhoeben: No need to duplicate the call, you can always do the try/catch and in the catch block decide whether you want to re-throw the exception or just set responseIsValid to false. Or am I missing something?

maartenjanvangool:
Oops, I see now I got some review comments. I'll take a look at it tomorrow :D

src/main/java/nl/hsac/fitnesse/fixture/util/HttpResponse.java
maartenjanvangool: I did not want to use (or touch) this method, because throwing a RuntimeException without specifing what kind of runtime exception seems like a pretty big smell to me. And catching *any* runtime exception and converting it to something specific does not make it better if you ask me.

src/main/java/nl/hsac/fitnesse/fixture/slim/HttpTest.java
maartenjanvangool: Yeah I have a tendency top press control alt L a lot... I'll revert these.

src/main/java/nl/hsac/fitnesse/fixture/util/XmlHttpResponse.java
```diff
@@ -22,7 +22,25 @@
...
         super.validResponse();
+        checkXmlResponse();
+    }
+
+    @Override
+    public boolean responseIsValid(boolean throwExceptionOnHttpRequestFailure) {
+        boolean responseIsValid = super.responseIsValid(throwExceptionOnHttpRequestFailure);
+        try {
+            checkXmlResponse();
+        } catch (RuntimeException e){
```
fhoeben:
```suggestion
        } catch (RuntimeException e) {
```

src/main/java/nl/hsac/fitnesse/fixture/util/XmlHttpResponse.java
```diff
@@ -22,7 +22,25 @@
...
+            if(throwExceptionOnHttpRequestFailure){
```
fhoeben:
```suggestion
            if (throwExceptionOnHttpRequestFailure) {
```

src/main/java/nl/hsac/fitnesse/fixture/util/HttpResponse.java
fhoeben: You are right about catching a general RuntimeException. I never used the 'slim' exception here because this code is also using from Fit fixtures, and Fit has no special handling for SlimFixtureException.

But we can probably to it the other way around. Define a custom subclass of RuntimeException (e.g. 'NonValidResponseReceivedException'). And have this thrown by the request methods. In the slim fixtures we can catch this and either convert it to 'false' or to a SlimFixtureException.
What are your thoughts on that?

maartenjanvangool:
Something like this?

maartenjanvangool:
thanks!

Issue #284, opened 2019-11-04
Fails to connect to android driver at GRID from versions above 4.5.5 


Taxanas:
Hello, Fried Hoeben,

I'm huge fan of your work, which makes lots of lives much easier (including mine :))

There's an issue with connectivity between hsac-fitnesse-fixtures and GRID with versions above 4.5.5, starting from this commit:

> https://github.com/fhoeben/hsac-fitnesse-fixtures/commit/8386eaed763950104b9387ea93a77a0f2aa6ae43

Im using hsac-fitnesse-appium method connectToAndroidDriverAtWithCapabilities:

In the end it requests for:

> {
  "desiredCapabilities": {
    "unhandledPromptBehavior": "ignore",
    "platformName": "Android",
    "unexpectedAlertBehaviour": "ignore"
  },
  "capabilities": {
    "alwaysMatch": {
      "platformName": "android",
      "unexpectedAlertBehaviour": "ignore",
      "unhandledPromptBehavior": "ignore"
    },
    "firstMatch": [
      {
      }

As i understand, "unexpectedAlertBehaviour" and "unhandledPromptBehavior" are not standard w3c caps and are not recognized by GRID (by versions above 2.53.3) and by appium.

4.5.5 version was requesting for:
> {
  "desiredCapabilities": {
    "platformName": "Android"
  },
  "capabilities": {
    "alwaysMatch": {
      "platformName": "android"
    },
    "firstMatch": [
      {
      }
    ]
  }
}

Any help or suggestion of workaround would be appreciated.

fhoeben:
I changed the unexepected  alert behavior because chrome's defaults changed.
I use this all the time with Selenium grid (3.141.59), without problem for chrome desktop.

I could image the setting is a problem for Appium, but for a normal grid I expect it to work. Are you experiencing trouble there also?

I have to admit I haven't looked at Appium for a while. Any chance you can create a PR to fix this for Appium (there is an acceptance test for desktop chrome handling of alerts, that should remain green ;-))?

Taxanas:
Thanks for fast response,

You are right, the problem with GRID only appears when using connectToAndroidDriverAtWithCapabilities method. 
While using connectToDriverAtWithCapabilities retrieves session without issues for same capabilities provided.
connectToDriverAtWithCapabilities json:
> {
  "desiredCapabilities": {
    "app": "*.apk",
    "appActivity": "*Activity",
    "appPackage": "*debug",
[...]

PR #285, opened 2019-11-05
add default chrome capabilities only for chrome. Fix https://github.c…

Taxanas:
…om/fhoeben/hsac-fitnesse-fixtures/issues/284

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverFactory.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverFactory.java
@@ -12,8 +12,12 @@ public interface DriverFactory {
     WebDriver createDriver();
 
     static void addDefaultCapabilities(MutableCapabilities capabilities) {
-        if (!capabilities.getCapabilityNames().contains(CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR)) {
-            capabilities.setCapability(CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR, UnexpectedAlertBehaviour.IGNORE);
+        if (capabilities.getCapabilityNames().contains(CapabilityType.BROWSER_NAME)) {
+            if (capabilities.getBrowserName().equals("chrome")){
+                if (!capabilities.getCapabilityNames().contains(CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR)) {
+                    capabilities.setCapability(CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR, UnexpectedAlertBehaviour.IGNORE);
+                }
+            }
         }
     }
 }
--- a/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverFactory.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverFactory.java
@@ -14,7 +14,7 @@ public interface DriverFactory {
     WebDriver createDriver();
 
     static void addDefaultCapabilities(MutableCapabilities capabilities) {
-        Set<String> capabilitieNames = capabilities.getCapabilityNames();
+        Set<String> capabilityNames = capabilities.getCapabilityNames();
         if (capabilitieNames.contains(CapabilityType.BROWSER_NAME)
                 && !capabilitieNames.contains(CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR)) {
             capabilities.setCapability(CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR, UnexpectedAlertBehaviour.IGNORE);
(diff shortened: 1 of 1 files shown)


fhoeben:
Does it make sense to invert the check: instead of only setting for chrome NOT set for Android (and maybe iOS, have you tested that)?
Since it is part of the W3C spec I would want to set it for all browsers, and not for those we find do not support it.

fhoeben:
Does it make sense to invert the check: instead of only setting for chrome NOT set for Android (and maybe iOS, have you tested that)?
Since it is part of the W3C spec I would want to set it for all browsers, and not for those we find do not support it.

Taxanas:
I agree, and will try it tomorrow 😉 
Actually, current Appium logs unknown cappabilities, ignores them, and works just fine.
My suggested PR is more of a workaround to connectToAndroidDriverAtWithCapabilities method adding those into desiredCapabilities and copying into capabilities in mentioned jsons. 
P. S. I got myself lost in debuging differences between connectToAndroidDriverAtWithCapabilities and connectToDriverAtWithCapabilities. 🙂

P. P. S. As i have an opportunity to communicate to you, might i ask whether there are plans for connectToWindowsDriverAtWithCapabilities? 🙂

fhoeben:
@tcnh didn't you once mention you created an extension to connect to Windows devices? Would that be something like the `connectToWindowsDriverAtWithCapabilities` @Taxanas is looking for? If so, any chance you can turn that into a PR (or did you already publish it yourself somewhere)?

Taxanas:
Hello, @fhoeben 
Wanted to ask your opinion on this:
https://github.com/fhoeben/hsac-fitnesse-fixtures/pull/285/commits/d4357146144a03370758023646b1b62a0470f0e3 
:)

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/DriverFactory.java
```diff
@@ -5,14 +5,18 @@
...
+
 /**
  * Interface to encapsulate creation of Selenium driver.
  */
 public interface DriverFactory {
     WebDriver createDriver();
 
     static void addDefaultCapabilities(MutableCapabilities capabilities) {
-        if (!capabilities.getCapabilityNames().contains(CapabilityType.UNEXPECTED_ALERT_BEHAVIOUR)) {
+        Set<String> capabilitieNames = capabilities.getCapabilityNames();
```
fhoeben:
```suggestion
        Set<String> capabilityNames = capabilities.getCapabilityNames();
```

tcnh:
I have to say I'm not completely up to date on appium right now. We decided to start an initiative to decouple windows fixture implementation from the mobile device fixtures, because the interaction is different.  I don't exactly know the curent status but I know some of our consultants are still effectively using https://repo.maven.apache.org/maven2/nl/praegus/toolchain-appium-fixtures/0.0.12/ (https://github.com/praegus/toolchain-appium-fixtures) for android, iOS and Windows 10 and have done some fixes on it recently, so maybe it contains what you need?

Taxanas:
Hello, @fhoeben ,

Today i tried connectToAndroidDriverAtWithCapabilities from praegus/toolchain-appium-fixtures. Same issue appears when trying to connect to grid.
@tcnh , thank you for sharing info :)

Any updates about pull request?

tcnh:
That's strange, I checked with a few colleagues that use toolchain-appium-fixtures.. none of them experience issues connecting with appium. Are you trying to get an appium session on a selenium grid? Because I think that will not work...? If that should work, I can look into what needs to change

Taxanas:
@tcnh ,

> Are you trying to get an appium session on a selenium grid?
Exactly ;) i have various browsers, os **and mobile devices** connected to grid via appium.

When i connect directly to node, everything is _ok_, but when i request for node via grid, it tries to parse json and fails :) 

> Because I think that will not work...?
I have already tested it ;) works smooth, **but**...  **unexpectedAlertBehaviour** and **unhandledPromptBehavior** should stay under desiredCapabilites, not in capabilities in json, that is passed to GRID when requiring for node, same way, that it is requested when using simple **connectToDriverAtWithCapabilities** from hsac fixtures **OR** preferably would not be added to json at all when not requiring browser to work with :)

More about it:
https://github.com/fhoeben/hsac-fitnesse-fixtures/issues/284

tcnh:
Ah, I get what the issue is now. 

Funny thing is that when using a local driver, addDefaultCapabilities is _only_ called for chrome. When using a remote driver, it's always called. That's not completely consistent (and some indication that these caps are not widely accepted?)

I think setting only for chrome is the sensible solution, but will that work when testing chrome _on_ an actual android device?

Taxanas:
@tcnh , that is correct ;)
If you ask for chrome browser on android, it would still be requested by _browserName_, right? ;)
Thats why i don't want to emit it by platformName ;)

Latest commit https://github.com/fhoeben/hsac-fitnesse-fixtures/pull/285/commits/bcc6fc55488e390654720a31a0e12f4ab004fdaf

Makes them being added whenever browser is requested from GRID :)

PR #293, opened 2020-01-01
Ensuring testDelete can run on its own

OrDTesters:
Currently, testDelete fails when run on its own, because the file to delete initially does not exist. This pull request copies the file into position first before deleting.

Please let me know if you want to discuss the changes more.

### DIFF
--- a/src/test/java/nl/hsac/fitnesse/fixture/slim/FileFixtureTest.java
+++ b/src/test/java/nl/hsac/fitnesse/fixture/slim/FileFixtureTest.java
@@ -144,7 +144,10 @@ public class FileFixtureTest {
     public void testDelete() {
         try {
             fixture.setDirectory(testResourcesDir);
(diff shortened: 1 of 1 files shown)


src/test/java/nl/hsac/fitnesse/fixture/slim/FileFixtureTest.java
```diff
@@ -144,7 +144,10 @@ private String getExpectedUrlPath() {
     public void testDelete() {
         try {
             fixture.setDirectory(testResourcesDir);
+            fixture.copyTo(txtFilename, copyFilename);
             fixture.delete(copyFilename);
+        } catch (IOException ioe) {
```
fhoeben: Please merge these two catch blocks into one. 
Best way probably is just to catch `Exception` instead of `SlimFixtureException`

PR #295, opened 2020-01-16
textOf feature for browsertest

lennyg1:
valueOf only works for specific types of elements. Sometimes it's required to capture the text of other types of elements, like buttons.

This is possible by using
|check|value of attribute|innerText|on|something|text| but this isn't very obvious and it's a lot uglier than text of something.


fhoeben:
From a technical perspective (when using technical selectors) I get the point. Is there also a reason to do this from an end-user perspective?

What I've always done in the past is either just wait for/ensure the text is visible, or (if it's a button or link) click <text>. That just seems much more natural than to check the text on an element: I either want to see the text or click on it. If the text is different the test will fail

lennyg1:
I want to know if the text is present and attached to a specific element (button in this case, could be other elements), not just anywhere on a screen.

I admit it's a bit of an edge case, but for an application that dynamically creates a page based on a template, it's an interesting way to check if the processing of said template has the expected outcome. I use it similarly to the tooltip for functionality.

I have an application that has different types of text based on other criteria and they have to be set properly and be present on the right elements.

fhoeben:
I feel its too much of an edge case to warrant making browser test even more complex. Especially since an easy workaround seems available: make a scenario that does exactly the same as the new method you propose.
```
|scenario|check text of _ is _ |place, text|
|check|value of attribute|innerText|on|@{place}|@{text}| 
```

Or if you have many such elements on a page you can even use it decision table style
```
|table template|check texts |
|$text=|value of attribute|innerText|on|@{place}|

|script|browser test|
|open|http://....|
...

|check texts|
|place|text?|
|somewhere|Button 1 text|
|someplace else|Button 2 text|
 
```

or do you see some big advantage for having specific methods for this?

lennyg1:
I personally have used a scenario similar to this, but over the years I've used it more and more and saw questions from people I work with related to capturing text from elements that value of simply doesn't capture. As it's not such an uncommon thing to do, I've included textOf in my own extended browser test.

The whole value of attribute innertext might be very obvious for you and me, but I've encountered a lot more users for whom it isn't as obvious and this has become a challenge in those cases. To me it's a similar concept as the tooltipOf function, that captures the title of an element.

An alternative might be to improve documentation for HSAC so these tips and tricks are part of the examples. However HSAC browsertest functionality is very descriptive, which means documentation isn't required for the most part and I'm afraid this would make things more complex.

fhoeben:
I see your point.

Given that `value of` already takes the element's text if no value attribute is present this really is only needed for buttons, correct? Or are there other elements that have both a value and text where testing the text is useful?

Taxanas:
> I want to know if the text is present and attached to a specific element (button in this case, could be other elements), not just anywhere on a screen.

Hello lennyg1,

Why not use:
|wait for visible |_text_  |in  |_element_  |

I use this for Ng and React elements and it works just fine, cause _element_ can be of any size.

lennyg1:
There are a few other HTML elements where this is possible. Admittedly those are pretty rare. Input, option, li, and of course button.

Might be too obscure to warrant the textOf function.

PR #307, opened 2020-02-13
Move browser window

violinner:
Added two commands to move the browser window on the desktop, with Home as the upper-left corner of the desktop, usage:

<code>
| script | browser test |
|| move browser down 100 and over 200 pixels: |
|set browser position x | 200 | y | 100 |
|| Move window to UL corner: |
|set browser position to home |
</code>

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
@@ -28,10 +28,13 @@ import org.openqa.selenium.By;
 import org.openqa.selenium.Cookie;
 import org.openqa.selenium.Dimension;
 import org.openqa.selenium.Keys;
+import org.openqa.selenium.Point;
 import org.openqa.selenium.SearchContext;
 import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.UnhandledAlertException;
 import org.openqa.selenium.WebDriver;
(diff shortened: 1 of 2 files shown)


src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -2321,6 +2321,14 @@ public void setBrowserSizeToMaximum() {
         getSeleniumHelper().setWindowSizeToMaximum();
     }
 
+    public void setBrowserPositionXY(int xPos, int yPos) {
+		getSeleniumHelper().setBrowserPositionXY(xPos, yPos);
+	}
+
+    public void setBrowserPositionToHome() {
```
fhoeben: Please remove this method. If a user wants that they should just make a scenario for it.

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
```diff
@@ -1148,6 +1149,13 @@ public void setWindowSizeToMaximum() {
         getWindow().maximize();
     }
 
+    public void setBrowserPositionXY(int xPos, int yPos) {
+		String currentWindowHandle = driver().getWindowHandle();
+		getTargetLocator().window(currentWindowHandle);
```
fhoeben: Are these first two lines needed? And if so do we reset the window to the previous value after changing the position?

PR #308, opened 2020-02-21
Windows auth in http test

tcnh:
Allow http client setup to configure http test to use windows credentials. This is convenient when services are coupled to domain accounts and authorization on endpoints is performed through Negotiate/NTLM methods. 

Only available on Windows offcourse and if used, the process running the tests must run under an authorized useraccount.

Also added a method to set username and password for basic auth in client setup, so that setting a basic auth header on requests is no longer needed.


fhoeben:
When basic credentials are set in this way can a user still override these using the explicit methods in http test? I.e. does one set a default here that can be overridden?
For windows authentication this is not possible (i.e. one has to setup the http client again/differently to make calls without it), correct?

I think the remark in this PR "Only available on Windows offcourse and if used, the process running the tests must run under an authorized useraccount." are good to add to the method's documentation wiki page

tcnh:
Docs updated.
To be clear: both settings can be overridden on a per request basis by setting the appropriate authorization header. 
If Windows auth is used, the same (windows) credentialsprovider will be used when authenticating to the proxy server.

PR #309, opened 2020-02-21
Ability to sign soap messages with xml http test

tcnh:
Added methods to xml http test to configure WS Security on SOAP requests. Supports Signature, Timestamp and Usernametoken.


wiki/FitNesseRoot/HsacExamples/SlimTests/HttpTests.wiki
```diff
@@ -101,17 +101,32 @@ The commands/keywords (i.e. public methods) added by !-nl.hsac.fitnesse.fixture.
...
+|repeat until xPath <expression> is <value>                                                |If 'expression' does not evaluates to 'value' for the last response: repeats last request until it does. Do not forget to set 'repeat at most times', or else the loop may run endlessly!            |
+|validate response against xsd <XSD Schema>                                                |Validates the response xml against the given XSD Schema                                                                                                                                              |
+|validate response against xsd file <XSD File>                                             |Validates the response xml against the XSD Schema in the given XSD File                                                                                                                              |
+|'''WS Security configuration options'''                                                                                                                                                                                                                                                         |
```
fhoeben: This sentence seems incomplete

src/main/java/nl/hsac/fitnesse/fixture/util/SoapSigningHelper.java
```diff
@@ -0,0 +1,248 @@
...
+import javax.xml.crypto.dsig.CanonicalizationMethod;
+import javax.xml.crypto.dsig.DigestMethod;
+import javax.xml.soap.*;
+import java.io.ByteArrayInputStream;
+import java.util.Arrays;
+import java.util.Properties;
+
+public class SoapSigningHelper {
+
+    public static final String VALID_DIGEST_ALGORITHMS = "SHA1, SHA256, SHA512, RIPEMD160";
```
fhoeben: Why not initialise these static fields with arrays (or lists), instead of splitting on each usage?

src/main/java/nl/hsac/fitnesse/fixture/slim/XmlHttpTest.java
```diff
@@ -173,4 +296,8 @@ public XmlHttpResponse getResponse() {
     protected XmlHttpResponse createResponse() {
         return new XmlHttpResponse();
     }
+
+    private SoapSigningHelper signingHelper() {
+        return Environment.getInstance().getSoapSigningHelper();
```
fhoeben: just use `getEnvironment()`

src/main/java/nl/hsac/fitnesse/fixture/slim/XmlHttpTest.java
```diff
@@ -173,4 +296,8 @@ public XmlHttpResponse getResponse() {
     protected XmlHttpResponse createResponse() {
         return new XmlHttpResponse();
     }
+
+    private SoapSigningHelper signingHelper() {
```
fhoeben: please make protected so custom subclasses can use it

src/main/java/nl/hsac/fitnesse/fixture/util/HttpResponse.java
```diff
@@ -51,6 +51,7 @@ public String getRequest() {
      * @param aRequest the request to set
      */
     public void setRequest(String aRequest) {
+        aRequest = Environment.getInstance().getSoapSigningHelper().signSoapMessageIfNeeded(aRequest);
```
fhoeben: I don't really like that the request is responsible for signing itself. Can this logic not be moved to XmlHttpTest?

src/main/java/nl/hsac/fitnesse/fixture/Environment.java
```diff
@@ -6,28 +6,7 @@
...
-import nl.hsac.fitnesse.fixture.util.PropertiesHelper;
-import nl.hsac.fitnesse.fixture.util.ReflectionHelper;
-import nl.hsac.fitnesse.fixture.util.SecretMasker;
-import nl.hsac.fitnesse.fixture.util.TextFormatter;
-import nl.hsac.fitnesse.fixture.util.TimeoutHelper;
-import nl.hsac.fitnesse.fixture.util.XMLFormatter;
-import nl.hsac.fitnesse.fixture.util.XPathHelper;
-import nl.hsac.fitnesse.fixture.util.XmlHttpResponse;
-import nl.hsac.fitnesse.fixture.util.ZipHelper;
+import nl.hsac.fitnesse.fixture.util.*;
```
fhoeben: Please leave the individual imports (I suggest you configure your editor to only use `*` imports if more than 999 imports are done from one package (i.e. never).

Star-imports can give nasty side effects if the packages you are using get more (or less) classes

tcnh:
> I like the feature in general.
> But I wonder whether this should be part of XmlHttpTest, adding quite a bit of complexity, or whether it would be better to create a separate (subclass) fixture e.g. `WsSecurityXmlHttpTest`.

I considered that, but as XML http test is mainly aimed at SOAP services I found it logical to integrate them (as wssec is a SOAP extension and introducing another fixture for interacting with SOAP services may be even less logical to the user?). If you think putting the wss methods in a separate subclass is a better solution, I'll be happy to move them.

wiki/FitNesseRoot/HsacExamples/SlimTests/HttpTests.wiki
tcnh: It was meant as a sub-header to visually separate the ws sec methods in the documentation. Renders like this: 
image

src/main/java/nl/hsac/fitnesse/fixture/util/SoapSigningHelper.java
tcnh: I knew this question was coming :-)
I use these constants in the javadoc comments, so that the context helper documentaion can show the possibilities (and needs no maintenance when the lists of supported options expands). As the only way to reference constants from javadoc is when the constants are immutable primitives or Strings. When using an array or list, the array itself is immutable, but the values inside are not and thus they are not shown in the javadoc output.

src/main/java/nl/hsac/fitnesse/fixture/util/HttpResponse.java
tcnh: I looked at that, but It would intoduce a lot of duplication (or refactoring) around the use of templates, as these are translated to the request content in the Environment and the Environment eventually sends the request (as it holds the client)

Moving the signing to XmlHttpTest would impy that we process the template there as well. What do you think?

wiki/FitNesseRoot/HsacExamples/SlimTests/HttpTests.wiki
```diff
@@ -101,17 +101,32 @@ The commands/keywords (i.e. public methods) added by !-nl.hsac.fitnesse.fixture.
...
+|ws sec apply soap signature before sending <true/false>                                   |Enable/disable appending a soap signature to the ws-sec header. This needs                                                                                                                           |
```
fhoeben: My previous comment  about the incomplete sentence was at the wrong line. I was referring to 
> ...header. This needs

wiki/FitNesseRoot/HsacExamples/SlimTests/HttpTests.wiki
fhoeben: A separation is nice indeed, that's why I thought a separate subclass, and table here, would be better.

src/main/java/nl/hsac/fitnesse/fixture/util/HttpResponse.java
fhoeben: I see your point. But I do believe the signing should not be a feature of HttpResponse. I see your point with the template handling in the environment. 
I'm not that happy anymore with the template handling (and calling the methods on HttpClient) being in the Environment . But there it is.

Maybe a custom subclass should have its own template handling, and just passing around the already signed result of a template. 
Another approach could be to add an extra (`Function<String,String>`) parameter to the methods in the environment that allows post processing of the template output. But this seems a bit over-the-top also.

Play around with it a bit. I think doing your own template handling is probably the most elegant solution

src/main/java/nl/hsac/fitnesse/fixture/util/SoapSigningHelper.java
```diff
@@ -0,0 +1,248 @@
...
+
+    /**
+     * Set keystore to use
+     *
+     * @param keyStore (wiki)file path to keystore file
+     * @param ksType   Type of keystore. Can be JKS, JCEKS or PKCS11
+     */
+    public void setKeyStore(String keyStore, String ksType) {
+        if (Arrays.stream(VALID_KEY_STORE_TYPES.split("\\s*,\\s*")).noneMatch(ksType.toUpperCase()::equals)) {
+            throw new SlimFixtureException(false, "Invalid keystore type: " + ksType.toUpperCase() + ". Valid options: " + Arrays.toString(VALID_KEY_STORE_TYPES.split("\\s*,\\s*")));
```
fhoeben:
```suggestion
            throw new SlimFixtureException(false, "Invalid keystore type: " + ksType.toUpperCase() + ". Valid options: " + VALID_KEY_STORE_TYPES);
```

src/main/java/nl/hsac/fitnesse/fixture/util/SoapSigningHelper.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    /**
+     * Set the identifier type.
+     *
+     * @param type Type name from WSS4J's WSCONSTANTS.
+     */
+    public void setKeyIdentifierType(String type) {
+        if (Arrays.stream(VALID_KEY_IDENTIFIER_TYPES.split("\\s*,\\s*")).noneMatch(type.toUpperCase()::equals)) {
+            throw new SlimFixtureException(false, "Invalid key identifier type: " + type.toUpperCase() + ". Valid options: " + Arrays.toString(VALID_KEY_IDENTIFIER_TYPES.split("\\s*,\\s*")));
```
fhoeben:
```suggestion
            throw new SlimFixtureException(false, "Invalid key identifier type: " + type.toUpperCase() + ". Valid options: " + VALID_KEY_IDENTIFIER_TYPES);
```

src/main/java/nl/hsac/fitnesse/fixture/util/SoapSigningHelper.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    /**
+     * Set the canonicalization
+     *
+     * @param cMethod Canonicalization Method to use.
+     */
+    public void setCanonicalizationMethod(String cMethod) {
+        if (Arrays.stream(VALID_CANONICALIZATION_METHODS.split("\\s*,\\s*")).noneMatch(cMethod.toUpperCase()::equals)) {
+            throw new SlimFixtureException(false, "Invalid canonicalization method: " + cMethod.toUpperCase() + ". Valid options: " + Arrays.toString(VALID_CANONICALIZATION_METHODS.split("\\s*,\\s*")));
```
fhoeben:
```suggestion
            throw new SlimFixtureException(false, "Invalid canonicalization method: " + cMethod.toUpperCase() + ". Valid options: " + VALID_CANONICALIZATION_METHODS);
```

src/main/java/nl/hsac/fitnesse/fixture/util/SoapSigningHelper.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    /**
+     * Set the digest algorithm to use
+     *
+     * @param digestMethod The method to use.
+     */
+    public void setDigestAlgorithm(String digestMethod) {
+        if (Arrays.stream(VALID_DIGEST_ALGORITHMS.split("\\s*,\\s*")).noneMatch(digestMethod.toUpperCase()::equals)) {
+            throw new SlimFixtureException(false, "Invalid digest method: " + digestMethod.toUpperCase() + ". Valid options: " + Arrays.toString(VALID_DIGEST_ALGORITHMS.split("\\s*,\\s*")));
```
fhoeben:
```suggestion
            throw new SlimFixtureException(false, "Invalid digest method: " + digestMethod.toUpperCase() + ". Valid options: " + VALID_DIGEST_ALGORITHMS);
```

src/main/java/nl/hsac/fitnesse/fixture/util/SoapSigningHelper.java
fhoeben: Quite a good reason ;-)
I don't believe you use the splits more than once in normal behaviour, so leave them as is, but maybe a comment to indicate why this is done?

wiki/FitNesseRoot/HsacExamples/SlimTests/HttpTests.wiki
tcnh: Agree, separated in WsSecurityHttpTest now

src/main/java/nl/hsac/fitnesse/fixture/util/HttpResponse.java
tcnh: I have put the WSSec methods in a separate xml http test extension now, that does its own template translation by overriding the necessary methods from http test. To do this, I had to change access for some fields on httptest from private to static. Could have used a getter, but it felt like overkill in this case. If you prefer using getters, i'll change them

fhoeben:
@tcnh I have to admit I didn't look at your recent changes yet but I feel I will not merge this PR. Although the extra feature is very nice it also makes that 'standalone.zip' al lot bigger (38 to 56MB).
I feel we should move this feature into its own project (like we did for email and pdf fixtures). So that people who want to use it can easily add it but other people get the smaller footprint.

tcnh:
That's understandable. I'll create a separate artifact for it and release that to maven central!

fhoeben:
Should we make some adjustments to HttpTest to make overriding template handling easier (I prefer instance methods over static ones)? Or will you go for a custom HttpResponse subclass that encrypts/signs its requests?

tcnh:
No need to alter httptest, I think I can extend on HttpTest and just override the template function to use the subclass' own template field.

PR #312, opened 2020-03-04
Full page screenshot

100000001:
I have replaced the Selenium screenshot taker with aShot, as Selenium does not seem to be able to take a full page screenshot.

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
@@ -1020,6 +1020,13 @@ public class SeleniumHelper<T extends WebElement> {
                 executeJavascript("arguments[0].scrollIntoView(" + scrollIntoViewArgs + ");", element);
     }
 
+    /**
+     * Takes screenshot of current page (as .png).
+     * @param baseName name for file created (without extension),
+     *                 if a file already exists with the supplied name an
+     *                 '_index' will be added.
+     * @return absolute path of file created.
+     */
     public String takeScreenshot(String baseName) {
         String result = null;
 
--- a/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
@@ -28,18 +28,15 @@ import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.Keys;
 import org.openqa.selenium.NoAlertPresentException;
 import org.openqa.selenium.NoSuchWindowException;
-import org.openqa.selenium.OutputType;
 import org.openqa.selenium.Platform;
 import org.openqa.selenium.SearchContext;
 import org.openqa.selenium.StaleElementReferenceException;
-import org.openqa.selenium.TakesScreenshot;
 import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.ie.InternetExplorerDriver;
 import org.openqa.selenium.interactions.Actions;
-import org.openqa.selenium.remote.Augmenter;
 import org.openqa.selenium.remote.RemoteWebDriver;
 import org.openqa.selenium.remote.ScreenshotException;
 import org.openqa.selenium.safari.SafariDriver;
@@ -47,8 +44,13 @@ import org.openqa.selenium.support.ui.ExpectedCondition;
 import org.openqa.selenium.support.ui.FluentWait;
 import org.openqa.selenium.support.ui.Select;
 import org.openqa.selenium.support.ui.WebDriverWait;
+import ru.yandex.qatools.ashot.AShot;
+import ru.yandex.qatools.ashot.Screenshot;
+import ru.yandex.qatools.ashot.shooting.ShootingStrategies;
 
 import java.io.IOException;
+import java.io.ByteArrayOutputStream;
+import javax.imageio.ImageIO;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.time.Duration;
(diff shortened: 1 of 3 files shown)


fhoeben:
With this PR will all screenshots always be full page (i.e. not just the viewport, what is visible, but the whole page)?
I do not believe that is what we should do. In general screenshots should capture what is on the screen, since that is what an end user would see and methods like 'is visible' also check what is on screen.

I do believe however that it would be a great feature to be able to take full page screenshots. But I would put that behind a separate method of BrowserTest. Just like one can check whether an element is visible, or visible somewhere on the page.

Additionally there could be a configuration option to switch the default screenshots, e.g. the ones taken on exceptions, between full page and viewport.

100000001:
All screenshots would be full page, yes.

I considered adding a 'takeFullPageScreenshot' method, but a comment in the Selenium screenshot code (TakeScreenshot.class) said:

> For WebDriver extending TakesScreenshot, this makes a best effort
> depending on the browser to return the following in order of preference:
> * Entire page
> * Current window
> * Visible portion of the current frame
> * The screenshot of the entire display containing the browser

So I thought this is already what Selenium is trying to do, but failing to do (at least for me in current Chrome) and this change sort of fixes this. 

I can move the new code to takeFullPageScreenshot and update the pull request, which will keep things backward compatible.

Another option might be to add an optional parameter to determine the screenshot size/type which then defaults to the current behaviour.
With an additional change this parameter could default to a global config value.

fhoeben:
I think a new method `fullPageScreenshot` would be best (I dropped 'take' since I expect the wiki to usually have 'show' in first column so the wiki content becomes something like: '|show|full page screenshot|order_page|'

100000001:
`fullPageScreenshot` is now a separate method in `BrowserTest` and in `SeleniumHelper`, the method is also added to the `BrowserTests` wiki.

I also updated a screenshot in the `HsacExamples.SlimTests.BrowserTests.ScriptTest` wiki to be full page. The test seems to be a bit fickle though, probably due to all the external scripts and content being loaded on the tested web pages.

fhoeben:
Yes those tests are a pain. PRs with more reliable, and instructive examples are welcome...

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -2001,13 +2001,43 @@ protected String savePageSource(String fileName, String linkText) {
...
+     * @param basename filename (below screenshot base directory).
+     * @return location of screenshot.
+     */
+    public String fullPageScreenshot(String basename) {
+        return takeScreenshotWith(
+            basename,
+            name -> 
+                getSeleniumHelper()
+                    .fullPageScreenshot(
+                        getScreenshotBasename(name)
```
fhoeben: Why is the extra `getScreenshotBasename` required here and not above?

Furthermore it seems a bit odd that both the basename is supplied to takeScreenshotWith, and all it does is pass it to the function. Seems easier if the caller would just place it inside the lambda (and then pass Supplier<String> instead of function)

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -2001,13 +2001,43 @@ protected String savePageSource(String fileName, String linkText) {
...
+        );
+    }
+
+    /**
+     * Takes screenshot using the passed maker function
+     * @param basename filename (below screenshot base directory).
+     * @param maker    function for taking the screenshot, should return the screenshot location.
+     * @return location of screenshot.
+     */
+    private String takeScreenshotWith(String baseName, Function<String, String> maker) {
```
fhoeben: Please make protected, so people can override in subclasses, should they want to

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
```diff
@@ -1019,25 +1024,77 @@ public void scrollTo(WebElement element, boolean toCenter) {
...
+     *                 if a file already exists with the supplied name an
+     *                 '_index' will be added.
+     * @param type     type of screenshot.
+     * @return absolute path of file created.
+     */
+    protected String takeScreenshot(String baseName, ScreenshotType type) {
+        String result = null;
+        byte[] png = null;
+
+        switch (type) {
```
fhoeben: Can this not just be two methods instead of switch on type (or at least make each branch just call a method)

fhoeben:
I did the cleanup I requested and pushed to a [branch](https://github.com/fhoeben/hsac-fitnesse-fixtures/tree/pull/312)

When [I tried to run the code](https://fhoeben.gitlab.io/-/hsac-fitnesse-fixtures/-/jobs/489397183/artifacts/example-results/example-test-4/HsacExamples.SlimTests.BrowserTests.ScriptTest.html) I repeatedly got an exception:

```
java.lang.RuntimeException: org.openqa.selenium.JavascriptException: javascript error: Cannot read property 'scrollHeight' of null
(Session info: chrome=80.0.3987.149)
Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:17:03'
System info: host: 'runner-72989761-project-17353740-concurrent-0', ip: '172.17.0.3', os.name: 'Linux', os.arch: 'amd64', os.version: '4.19.78-coreos', java.version: '1.8.0_242'
Driver info: org.openqa.selenium.remote.RemoteWebDriver
Capabilities {acceptInsecureCerts: false, browserName: chrome, browserVersion: 80.0.3987.149, chrome: {chromedriverVersion: 80.0.3987.106 (f68069574609..., userDataDir: /tmp/.com.google.Chrome.bGmdto}, goog:chromeOptions: {debuggerAddress: localhost:37881}, javascriptEnabled: true, networkConnectionEnabled: false, pageLoadStrategy: normal, platform: LINUX, platformName: LINUX, proxy: Proxy(), setWindowRect: true, strictFileInteractability: false, timeouts: {implicit: 0, pageLoad: 300000, script: 30000}, unhandledPromptBehavior: ignore, webdriver.remote.sessionid: 4b289f0c92c2d703edab6193ba1...}
Session ID: 4b289f0c92c2d703edab6193ba154e51
at ru.yandex.qatools.ashot.util.InnerScript.execute(InnerScript.java:32)
at ru.yandex.qatools.ashot.shooting.ViewportPastingDecorator.getFullHeight(ViewportPastingDecorator.java:67)
at ru.yandex.qatools.ashot.shooting.ViewportPastingDecorator.getScreenshot(ViewportPastingDecorator.java:41)
at ru.yandex.qatools.ashot.shooting.ViewportPastingDecorator.getScreenshot(ViewportPastingDecorator.java:35)
at ru.yandex.qatools.ashot.AShot.takeScreenshot(AShot.java:145)
at nl.hsac.fitnesse.fixture.util.selenium.SeleniumHelper.fullPageScreenshot(SeleniumHelper.java:1077)
at nl.hsac.fitnesse.fixture.slim.web.BrowserTest.createFullPageScreenshot(BrowserTest.java:2077)
```

PR #314, opened 2020-03-26
Remove nested map keys

tcnh:
Support for removal of nested keys in map fixture.

It was possible to set a value like map.submap.key, but we could not remove using the same notation where periods are expanded to nested maps.

This PR allows to use clear value with a nested map path using periods.


src/main/java/nl/hsac/fitnesse/fixture/slim/SlimFixtureWithMap.java
```diff
@@ -124,9 +124,8 @@ protected Base64Fixture getBase64Fixture() {
      * @return true if value was present.
      */
     public boolean clearValue(String name) {
-        String cleanName = cleanupValue(name);
-        boolean result = getCurrentValues().containsKey(cleanName);
-        getCurrentValues().remove(cleanName);
+        boolean result = getMapHelper().getValue(getCurrentValues(), name) != null;
```
fhoeben: this check is not quite the same as the old result value. A map may contain a key with a `null` value, in that case `containsKey()` would return `true` and the new code `false`.
I don't think it's very important, but it is not backward compatible. Any chance you could update the implementation to make it fully backward compatible (maybe by having `removeFrom()` return a boolean to indicate whether the key was present?

tcnh: Agree, pushed a commit that ensures it only returns false if the key doesn't exist

PR #328, opened 2020-06-17
added methods to support dragging and dropping of React elements

Taxanas:
Hello @fhoeben,

Experienced strange behavior with react elements. They need to be dragged differently than standard elements.
Added new methods to support that. Will wait for your comments.

P.s. tried adding some example tests, but didn't find any example website.


fhoeben:
Thanks for figuring this out. Sorry I did not respond quicker. 
The lack of an acceptance test bothers me a bit. Any chance you can create a small sample page, or use one of the samples at https://reactjsexample.com/tag/drag-drop/ or https://react.rocks/tag/Drag_Drop?

Taxanas:
Hello, @fhoeben 

Thanks for the suggested https://react.rocks/tag/Drag_Drop?
It's good you were not quick to accept the PR. 

I have read more about https://github.com/clauderic/react-sortable-hoc

According documentation:
`Click events being swallowed
By default, react-sortable-hoc is triggered immediately on mousedown. If you'd like to prevent this behaviour, there are a number of strategies readily available. You can use the distance prop to set a minimum distance (in pixels) to be dragged before sorting is enabled. You can also use the pressDelay prop to add a delay before sorting is enabled. Alternatively, you can also use the SortableHandle HOC.`
image

Therefore, i need to expand this PR, to make it more universal and work in other users cases as well as mine :)

Taxanas:
Hello again, @fhoeben ,

This PR includes:
* 4 new methods to support https://github.com/clauderic/react-sortable-hoc limitations
* acceptance tests for those
* BrowserTest documentation extension 

Please check when you can

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
```diff
@@ -600,6 +600,38 @@ public void dragAndDropToOffsetXY(WebElement element, Integer xOffset, Integer y
...
+    /**
+     * Simulates a drag from source react element and drop to target element
+     * Performs a slight user defined delay before drag to avoid limitations of "pressDelay"
+     * More info: https://github.com/clauderic/react-sortable-hoc
+     * @param source element to start the drag
+     * @param delay long to define how much to delay in millis after clickAndHold
+     * @param xOffset horizontal integer offset destination for dropping (calculated from given element center).
+     * @param yOffset vertical integer offset destination for dropping (calculated from given element center).
+     */
+    public void dragWithDelayAndDropToOffsetXY (WebElement source, long delay, Integer xOffset, Integer yOffset) {
```
fhoeben, line 628:
```suggestion
    public void dragWithDelayAndDropToOffsetXY(WebElement source, long delay, Integer xOffset, Integer yOffset) {
```

Taxanas: Thanks for response,

I agree with your ideas to be configurable and use one method, but the problem i see is backwards compatibility for the ones, that already use html5DragAndDrop (if we merge it with dragAndDrop).
Also, I believe there can be projects that have different UI elements which work with limitations and the ones, that work with simple drag and drop.
From the selenium actions perspective, both elements should work with drag and drop used with limitation, as drag and drop with limitation only adds additional actions (moveByOffset or pause), but html5draganddrop is different story.

My suggestion
Leave 2 methods:
DragAndDropTo (would try/return simple or distance or delay)
Html5dragAndDrop

Will think of something and commit.

Taxanas:
Hello, @fhoeben , changes done
Please check when you can

PR #330, opened 2020-06-26
Configurable multipart boundary

tcnh:
Added POST and PUT file methods to http test that allow to specify the multipart boundary string to use when constructing the multipart form data.


fhoeben:
Seems fine at first glance. 
Out of curiosity: in what situations did the need to configure the boundary arise?

tcnh:
We have a project where the service requires the multipart boundary to be specified in the content-type header. As I found no way to get the generated boundary without parsing the whole request, I chose to implement it the other  way around. As a bonus, the test requests can now be easily identified in the service logs because they have readable strings as boundary.

fhoeben:
I see. Should we maybe go one step further and allow a single call on HttpTest to set both the header (including boundary) and ensure that it is sent correctly? I'm thinking that the header `Content-Type` should probably be `multipart/form-data; boundary=<boundary value>`.

tcnh:
Good idea, I'll add that later today

tcnh:
Testing some more with this, It seems the http client already silently sets the boundary in the content-type header. That would render this whole change unnecessary. Will check with the test environment on monday why messages with a custom boundary were accepted and with a generated one were not.

tcnh:
Closing this. HttpClient implicitly sets the boundary to the content-type header of it's not explicitly set to a value without the boundary (as was the case in our situation). Renders this functionality useless.

PR #333, opened 2020-07-07
Add ability to screenshot only a given element

tcnh:
This PR adds a method (and its -in variant) to create an image of a specific element (scrolling the element into view and cropping the screenshot to the element boundaries)

Our use case is providing specific visual evidence in test reports, as well as generating specific images to be used in documentation.


src/main/java/nl/hsac/fitnesse/fixture/util/selenium/SeleniumHelper.java
```diff
@@ -1057,6 +1063,44 @@ public String takeScreenshot(String baseName) {
...
+        String result = null;
+
+        WebDriver d = driver();
+
+        if (!(d instanceof TakesScreenshot)) {
+            d = new Augmenter().augment(d);
+        }
+        if (d instanceof TakesScreenshot) {
+            TakesScreenshot ts = (TakesScreenshot) d;
+            File screenshot = ts.getScreenshotAs(OutputType.FILE);
```
fhoeben, line 1084: This first part seems to share logic with the existing screenshot functionality. Can you see whether that can be extracted to reduce code/functionality duplication?

tcnh: I extracted the augmenting and casting to a separate method that returns a TakesScreenshot instance. Let me know if that helps.

PR #335, opened 2020-07-14
Chromium based edgedriver

timovd:
Add support for new (chromium based) edgedriver


fhoeben:
Unfortunately upgrading the webdriverextensions-maven-plugin has broken my normal workflow see https://github.com/webdriverextensions/webdriverextensions-maven-plugin/issues/49

@timovd any chance you see a way for this change to work with the older version of the plugin?

PR #336, opened 2020-07-15
update to latest chromedriver 84.0.4147.30

timovd:
Add latest chromedriver 84.0.4147.30


fhoeben:
Is there a 84 based Edge already?

timovd:
> Is there a 84 based Edge already?

Apparently there's 86 already (https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver/), but honestly I haven't looked into this. And I don't know which is a stable release.

timovd:
> Is there a 84 based Edge already?

MS just released 84.0.522.40

PR #337, opened 2020-07-16
Revert webdriverextension

timovd:
See https://github.com/webdriverextensions/webdriverextensions-maven-plugin/issues/49


fhoeben:
Does the edge download still work? I thought specifying the file to get was also a new thing in the latest version of the plugin.

timovd:
@fhoeben https://github.com/webdriverextensions/webdriverextensions-maven-plugin/issues/49 looks like this is fixed. Would it be useful to get v3.3.0 into the fixtures, since there are some useful improvements and fixes?

PR #343, opened 2020-09-07
empty elementOfJsonPath

vegterb:
return null with empty elementOfJsonPath

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/JsonFixture.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/JsonFixture.java
@@ -73,7 +73,7 @@ public class JsonFixture extends SlimFixture {
 
     public Object elementOfJsonPath(int index, String path) {
         List<Object> all = listJsonPathMatches(path);
-        return all.size() > 0 ? all.get(index) : null;
+        return all.size() > index ? all.get(index) : null;
     }
 
     public int jsonPathCount(String path) {
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/JsonFixture.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/JsonFixture.java
@@ -73,7 +73,7 @@ public class JsonFixture extends SlimFixture {
 
     public Object elementOfJsonPath(int index, String path) {
         List<Object> all = listJsonPathMatches(path);
-        return all.get(index);
+        return all.size() > 0 ? all.get(index) : null;
     }
 
     public int jsonPathCount(String path) {
--- a/src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
+++ b/src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
@@ -68,7 +68,7 @@ public class JsonFixtureTest {
     public void testElementOfJsonPathResults(){
         JsonFixture cont = new JsonFixture();
         cont.load(JSON);
-        Object result = cont.elementOfJsonPath(0,jsonPathWithResults);
+        Object result = cont.elementOfJsonPath(0, "$.store.book[?(@.category==\"fiction\")].author");
         assertEquals(expectedResult,result);
     }
 }
--- a/src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
+++ b/src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
@@ -2,20 +2,11 @@ package nl.hsac.fitnesse.fixture.slim;
 
 import org.junit.Test;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 
 public class JsonFixtureTest {
     private JsonFixture fixture = new JsonFixture();
-    private final String jsonPathNoResults = "$.store.book[?(@.category==\"test\")].author";
-    private final String jsonPathWithResults = "$.store.book[?(@.category==\"fiction\")].author";
-    private final String expectedResult = "Evelyn Waugh";
-
 
     private static String JSON =
             "{\n" +
@@ -58,17 +49,18 @@ public class JsonFixtureTest {
 
     @Test
     public void testElementOfJsonPathNoResult() {
-        JsonFixture cont = new JsonFixture();
-        cont.load(JSON);
-        Object result = cont.elementOfJsonPath(0, jsonPathNoResults);
+        fixture.load(JSON);
+        Object result = fixture.elementOfJsonPath(0, "$.store.book[?(@.category==\"test\")].author");
+        assertNull(result);
+
+        result = fixture.elementOfJsonPath(3, "$.store.book[?(@.category==\"fiction\")].author");
         assertNull(result);
     }
 
     @Test
     public void testElementOfJsonPathResults(){
-        JsonFixture cont = new JsonFixture();
-        cont.load(JSON);
-        Object result = cont.elementOfJsonPath(0, "$.store.book[?(@.category==\"fiction\")].author");
-        assertEquals(expectedResult, "Evelyn Waugh");
+        fixture.load(JSON);
+        Object result = fixture.elementOfJsonPath(0, "$.store.book[?(@.category==\"fiction\")].author");
+        assertEquals("Evelyn Waugh", result);
     }
 }
--- a/src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
+++ b/src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
@@ -1,0 +1,76 @@
+package nl.hsac.fitnesse.fixture.slim;
+
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class JsonFixtureTest {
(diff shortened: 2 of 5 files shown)


fhoeben:
Looks good. Would you mind adding a test?

vegterb:
JsonFixtureTest Class & twee testen aangemaakt

src/main/java/nl/hsac/fitnesse/fixture/slim/JsonFixture.java
```diff
@@ -73,7 +73,7 @@ public String createFileFromBase64ContentOf(String baseName, String jsonPath) {
 
     public Object elementOfJsonPath(int index, String path) {
         List<Object> all = listJsonPathMatches(path);
-        return all.get(index);
+        return all.size() > 0 ? all.get(index) : null;
```
fhoeben, line 76:
```suggestion
        return all.size() > index ? all.get(index) : null;
```

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
```diff
@@ -0,0 +1,74 @@
...
+                    "            \"color\": \"red\",\n" +
+                    "            \"price\": 19.95\n" +
+                    "        }\n" +
+                    "    },\n" +
+                    "    \"expensive\": 10\n" +
+                    "}\n";
+
+    @Test
+    public void testElementOfJsonPathNoResult() {
+        JsonFixture cont = new JsonFixture();
```
fhoeben, line 61: Why not just use `fixture`?

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
```diff
@@ -0,0 +1,74 @@
...
+        cont.load(JSON);
+        Object result = cont.elementOfJsonPath(0,jsonPathNoResults);
+        assertNull(result);
+    }
+
+    @Test
+    public void testElementOfJsonPathResults(){
+        JsonFixture cont = new JsonFixture();
```
fhoeben, line 69: Why not just use `fixture`?

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
```diff
@@ -0,0 +1,74 @@
...
+                    "        }\n" +
+                    "    },\n" +
+                    "    \"expensive\": 10\n" +
+                    "}\n";
+
+    @Test
+    public void testElementOfJsonPathNoResult() {
+        JsonFixture cont = new JsonFixture();
+        cont.load(JSON);
+        Object result = cont.elementOfJsonPath(0,jsonPathNoResults);
```
fhoeben, line 63:
```suggestion
        Object result = cont.elementOfJsonPath(0, jsonPathNoResults);
```

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertNull(result);
+    }
+
+    @Test
+    public void testElementOfJsonPathResults(){
+        JsonFixture cont = new JsonFixture();
+        cont.load(JSON);
+        Object result = cont.elementOfJsonPath(0,jsonPathWithResults);
```
fhoeben, line 71:
```suggestion
        Object result = cont.elementOfJsonPath(0, "$.store.book[?(@.category==\"fiction\")].author");
```

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertEquals(expectedResult,result);
```
fhoeben, line 72:
```suggestion
        assertEquals(expectedResult, "Evelyn Waugh");
```

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.stream.Stream;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class JsonFixtureTest {
+    private JsonFixture fixture = new JsonFixture();
+    private final String jsonPathNoResults = "$.store.book[?(@.category==\"test\")].author";
+    private final String jsonPathWithResults = "$.store.book[?(@.category==\"fiction\")].author";
+    private final String expectedResult = "Evelyn Waugh";
```
fhoeben, line 17: remove

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class JsonFixtureTest {
+    private JsonFixture fixture = new JsonFixture();
+    private final String jsonPathNoResults = "$.store.book[?(@.category==\"test\")].author";
+    private final String jsonPathWithResults = "$.store.book[?(@.category==\"fiction\")].author";
```
fhoeben, line 16: remove

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
```diff
@@ -0,0 +1,74 @@
...
+                    "    },\n" +
+                    "    \"expensive\": 10\n" +
+                    "}\n";
+
+    @Test
+    public void testElementOfJsonPathNoResult() {
+        JsonFixture cont = new JsonFixture();
+        cont.load(JSON);
+        Object result = cont.elementOfJsonPath(0,jsonPathNoResults);
+        assertNull(result);
```
fhoeben, line 64: Add test that getting element 1 also returns `null`

fhoeben (approved):
Looks good. A few minor points to finish it I believe

src/main/java/nl/hsac/fitnesse/fixture/slim/JsonFixture.java
vegterb: Ah, yes... Interesting. This would make sure that when you chose an element which does not exist.

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
vegterb: Interesting, why would you keep these values in code and not variables? Does this improve speed? It looked cleaner with the variable, but either way is fine with me.

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
vegterb: Blindly added, but I expect you wanted to replace the expectedResult variable here.

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
vegterb: Agreed

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonFixtureTest.java
vegterb: Add test with no result, where the jsonPath results in 3 items, but item 3 returns null.

vegterb:
@fhoeben changes made, thanks for the feedback.

fhoeben:
Just remembered `JsonHttpTest`: has its own handling (if only Java had mixins) of getting elements. Please make the same change there. 
Sorry for not mentioning it earlier...

vegterb:
@fhoeben as per your comment I updated `JsonHttpTest` but got a bit stuck on the unit test, my girlfriend suggested Mockito, which was already part of your project so I also updated the JsonFixtureTest. Let me know what you think.

fhoeben:
Although I really like mockito, I try to keep away from using spies as much as possible as they tend to couple your test too tightly to the implementation. Instead I suggest you take a look at wiki/FitNesseRoot/HsacAcceptanceTests/SlimTests/HttpTest/JsonHttpTest.wiki line 30, I believe adding one line there to check element 2 would be the easiest way to test the changed behaviour (line 30 tests the behaviour when the element _is_ present)

vegterb:
I could of course not use the spy implementation and just use a mock for the `HttpResponse` class and keep using all the methodes within the 'JsonHttpTest' class. It's you're call... mock or revert and use:
`|check        |element              |2                   |of json path|airplanes[?(@.airline=='KLM')].type|null|`
is also no problem.

I will revert the changes in the 'JsonFixtureTest'.

fhoeben:
I would prefer `|check |element |2 |of json path|airplanes[?(@.airline=='KLM')].type|null|` 

The spy indeed isolates the specific method, but that method is the implementation detail. The FitNesse test checks the behaviour of the class given a certain response from the remote server, it's completely decoupled from how the fixture arrives at that response. So JsonHttpTest could be completely refactored (for instance by delegating to JsonFixture, or even rewriting it in Kotlin) and the test needs no change: it still verifies whether the correct response is given.

vegterb:
@fhoeben all changes made, should be good now.

src/test/java/nl/hsac/fitnesse/fixture/slim/JsonHttpTestTest.java
```diff
@@ -6,8 +6,7 @@
```
fhoeben, line 6: Please revert all changes to this file (`*` imports are a bad idea, and I like newlines at the end of my files)

fhoeben (changes requested):
Looks good. One minor thing left

PR #345, opened 2020-09-15
Support map of capabilities for (chromium) edge, like chrome

tcnh:
Since edge is now chromium based, we can also pass it a profile map to better parameterize it. (i.e. run headless, start-maximized set screen resolution, etc)

This PR ensures that when a a map is passed to startDriverForWithProfile and msedge is requested, that the profile map is applied as capabilities. (ms:edgeOptions capability)

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/LocalDriverFactory.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/LocalDriverFactory.java
@@ -1,12 +1,10 @@
 package nl.hsac.fitnesse.fixture.util.selenium.driverfactory;
 
-import com.galenframework.generator.raycast.EdgesContainer;
 import nl.hsac.fitnesse.fixture.slim.SlimFixtureException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.chrome.ChromeDriver;
 import org.openqa.selenium.chrome.ChromeOptions;
 import org.openqa.selenium.edge.EdgeDriver;
-import org.openqa.selenium.edge.EdgeOptions;
 import org.openqa.selenium.firefox.FirefoxDriver;
 import org.openqa.selenium.firefox.FirefoxOptions;
 import org.openqa.selenium.firefox.FirefoxProfile;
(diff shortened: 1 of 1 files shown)


src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/LocalDriverFactory.java
```diff
@@ -49,6 +50,9 @@ public WebDriver createDriver() {
             } else if ("internetexplorerdriver".equalsIgnoreCase(driverClass.getSimpleName())) {
                 InternetExplorerOptions ieOptions = getInternetExplorerOptions(profile);
                 driver = new InternetExplorerDriver(ieOptions);
+            } else if ("edgedriver".equalsIgnoreCase(driverClass.getSimpleName())) {
```
fhoeben, line 53: This will add another deprecation warning to the build. Do you see how hard/easy it would be to switch to EdgeOptions (and ChromeOptions). I would appreciate another PR for that...

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/LocalDriverFactory.java
```diff
@@ -132,4 +136,12 @@ public static InternetExplorerOptions getInternetExplorerOptions(Map<String, Obj
         }
         return ieOptions;
     }
+
+    public static DesiredCapabilities getChromiumEdgeOptions(Map<String, Object> profile) {
+        DesiredCapabilities capabilities = DesiredCapabilities.edge();
+        if(profile != null) {
```
fhoeben, line 142:
```suggestion
        if (profile != null) {
```

fhoeben (approved):
Minor issue

src/main/java/nl/hsac/fitnesse/fixture/util/selenium/driverfactory/LocalDriverFactory.java
tcnh: I had a quick look, but I didn't see an easy solution. I hereby promise to look into it and do a new PR somewhere later this month :-)

However we have a current need for chromium capabilities in the chromium version of edge, so if this could act as an intermediate solutions that would be very helpful

PR #359, opened 2020-11-13
Allow using plaintext http://files urls

tcnh:
Previously, http://files url's referencing the wiki's file section were not translated to an actual file path when they were not translated to a html link first when used with i.e. FileFixture.

This PR allows to use http://files references from a literal table (`!|table|`)


fhoeben:
Can you maybe add a test (wiki page)?

tcnh:
I added an example to FileFixture's example page. I think this change is not big enough to warrant it's own dedicated example/acceptance test page :-)

PR #360, opened 2020-11-15
Fix Nondeterministic Ordering in Tests

chrnndz3:
The following 10 tests:
`nl.hsac.fitnesse.fixture.slim.JsonHttpTestTest#testPutValuesAsJson`
`nl.hsac.fitnesse.fixture.slim.JsonHttpTestTest#testPostValuesAsJson`
`nl.hsac.fitnesse.fixture.slim.JsonHttpTestTest#testDeleteValuesAsJson`

`nl.hsac.fitnesse.fixture.slim.JsonHttpTestTest#testFormatJson`
`nl.hsac.fitnesse.fixture.slim.JsonHttpTestTest#testFormatJsonArray`
`nl.hsac.fitnesse.fixture.util.JsonHelperTest#testFormatSimple`

`nl.hsac.fitnesse.fixture.util.JsonHelperTest#testSortingOnNumber` 
`nl.hsac.fitnesse.fixture.util.JsonHelperTest#testSortingOnString`

`nl.hsac.fitnesse.fixture.util.XmlHttpResponseTest#testCheckXPathsIncorrect`

`nl.hsac.fitnesse.fixture.util.NamespaceContextImplTest#testAddMultipleGet`

are flaky tests. The first 3 tests depend on a function called `jsonEncodeCurrentValues()` in `JsonHttpTest.java` . While debugging, we found out that `JSONObject` was changing the order of the `LinkedHashMap`. As a result, to guarantee the insertion order, the `Gson` library was used instead of `JSONObject`. This library will now convert the ordered map into an ordered string.
More details can be read here:
https://stackoverflow.com/questions/29491281/building-ordered-json-string-from-linkedhashmap

The next 3 tests depend on a function called format() in JsonHelper.java. While debugging, we found out that JSONObject was changing the order of the LinkedHashMap. As a result, to guarantee the order, the `Gson` library was used instead of `JSONObject`. 

The next 2 tests depend on two functions called sort() and format() in JsonHelper.class. While debugging, we found out these tests call `sort()` which calls `sort()` (but with different parameters ) this function calls -> `convertToArray()` then  `jsonStringToMap()` -> `jsonObjectToMap()`. Therefore, during this flow it calls JSONObject  a few times which changed the order object. As a result, instead of `org.junit.Assert.assertEquals` we used `org.skyscreamer.jsonassert.JSONAssert`. 
Skyscreamer's JSONAssert is library which compares jsons but the ordering doesn't matter if the _strict mode_ is false which in this case it is. 
More details can be read here:
https://github.com/skyscreamer/JSONassert
https://stackoverflow.com/questions/2253750/testing-two-json-objects-for-equality-ignoring-child-order-in-java

The 2nd to last flaky test called `testCheckXPathsIncorrect()` in  `XmlHttpResponseTest.java` depends on a helper method called `createExprToCheck()` which creates a HashMap<String, String>.  However,  the order of `amount`, `result`, `unknownKey`,  and `status` is not guaranteed. To guarantee insertion order, the HashMap is changed to a LinkedHashMap in createExprToCheck() which is the root cause of flakiness. 

Lasty, the last flaky tests `testAddMultipleGet` in `NamespaceContextImplTest.java` also didn't gaurantee order. Therefore, changed `HashMap` to `LinkedHashMap` in class `NamespaceContextImpl.java` which initializes a HashMap and where the root cause is for nondeterministic ordering.

Build and unit tests pass:
Screen Shot 2020-11-15 at 10 53 15 PM

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/util/JsonHelper.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/util/JsonHelper.java
@@ -1,6 +1,8 @@
 package nl.hsac.fitnesse.fixture.util;
 
-import com.google.gson.*;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonParser;
 import net.minidev.json.JSONArray;
 import nl.hsac.fitnesse.fixture.Environment;
 import org.apache.commons.lang3.StringUtils;
@@ -26,26 +28,15 @@ public class JsonHelper implements Formatter {
     public String format(String json) {
         String result = null;
         if (json != null){
+            Gson gson = new GsonBuilder().setPrettyPrinting().create();
+            JsonParser jsonParser = new JsonParser();
+
             if (json.startsWith("{")) {
-                // Fixed JsonHttpTestTest#testFormatJson
-                return new Gson().toJson(json);
+                return gson.toJson(jsonParser.parse(json).getAsJsonObject());
             } else if (json.startsWith("[")) {
-//                JSONObject jsonObject = new JSONObject("{'a': " + json + "}");
-//                org.json.JSONArray array = (org.json.JSONArray) jsonObject.get("a");
-//                result = array.toString(4);
-
-                JsonParser jsonParser = new JsonParser();
-                Gson gson = new GsonBuilder().setPrettyPrinting().create();
-                String prettyJson = gson.toJson(jsonParser.parse(json).getAsJsonArray().get(0).getAsJsonObject());
-                System.out.println( "json prettify"+ prettyJson);
-
-                org.json.JSONArray array = new org.json.JSONArray("[" + prettyJson + "]");
-                System.out.println(gson.toJson(array));
-
-                result = array.toString(4);
-
-//                JsonArray jsonArray1 = (JsonArray) new JsonParser().parse(json).getAsJsonArray();
-//                org.json.JSONArray array = (org.json.JSONArray) jsonObject.get("a");
+                String prettyJson;
+                prettyJson = gson.toJson(jsonParser.parse(json).getAsJsonArray());
+                result = prettyJson;
             }
         }
         return result;
--- a/src/main/java/nl/hsac/fitnesse/fixture/util/JsonHelper.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/util/JsonHelper.java
@@ -32,7 +32,7 @@ public class JsonHelper implements Formatter {
             JsonParser jsonParser = new JsonParser();
 
             if (json.startsWith("{")) {
-                return gson.toJson(jsonParser.parse(json).getAsJsonObject());
+                result = gson.toJson(jsonParser.parse(json).getAsJsonObject());
             } else if (json.startsWith("[")) {
                 result = gson.toJson(jsonParser.parse(json).getAsJsonArray());
             }
--- a/src/main/java/nl/hsac/fitnesse/fixture/util/JsonHelper.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/util/JsonHelper.java
@@ -1,5 +1,6 @@
 package nl.hsac.fitnesse.fixture.util;
 
+import com.google.gson.*;
 import net.minidev.json.JSONArray;
 import nl.hsac.fitnesse.fixture.Environment;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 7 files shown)


src/main/java/nl/hsac/fitnesse/fixture/util/JsonHelper.java
```diff
@@ -25,12 +28,13 @@
     public String format(String json) {
         String result = null;
         if (json != null){
+            Gson gson = new GsonBuilder().setPrettyPrinting().create();
+            JsonParser jsonParser = new JsonParser();
+
             if (json.startsWith("{")) {
-                result = new JSONObject(json).toString(4);
+                return gson.toJson(jsonParser.parse(json).getAsJsonObject());
```
fhoeben, line 35: Please don't return from `if` branch and reduce code duplication between branches

src/test/java/nl/hsac/fitnesse/fixture/util/JsonHelperTest.java
```diff
@@ -89,20 +90,20 @@ public void testSortingOnNumber() {
...
-                        "        {\n" +
-                        "            \"price\": 8.95,\n" +
-                        "            \"category\": \"reference\"\n" +
-                        "        },\n" +
-                        "        {\n" +
-                        "            \"price\": 18.95,\n" +
-                        "            \"category\": \"areference\"\n" +
-                        "        }\n" +
-                        "    ]\n" +
+        JSONAssert.assertEquals("{\n" +
```
fhoeben, line 93: Why do we need to use JSONAssert here, and especially with `strict=false`? I thought the whole point of this PR was that the JSON was completely stable?

fhoeben (commented):
Looks pretty good. I only don't get why JSONAssert is needed, as all the other changes should make the JSON stable, correct?

src/test/java/nl/hsac/fitnesse/fixture/util/JsonHelperTest.java
chrnndz3: @fhoeben JSONAssert has two strict modes. If you set the strict mode to `false`,  then this means that compare the two jsons but the order doesn't matter. If it's true, then it will compare with two jsons while checking for the order. 

For the tests `SortingOnNumber` and `testSortingOnString`, the tests are coming out of order each time. When debugging there were changes to be done but to avoid a substantial amount of change. I decided to introduce JSONAssert which does assertEquals but doesn't change the order but everything else is checked. 

E.g. of strict mode = false
Json 1: 
``` json
{
  "firstName": "Billy",
  "lastName": "Sandy"
} 
```

Json 2 (different order):
``` json
{
  "lastName": "Sandy",
  "firstName": "Billy"
} 
```

JSONAssert will compare these two json but the order won't matter as long as long as everything else is the same.
Here's the link to the library: https://github.com/skyscreamer/JSONassert.

fhoeben:
I never had any problems with the ordering of the results in my jUnit tests, or acceptance tests I wrote with this helpers (not locally, nor on CI servers). On what platform and with which JVM did you encounter the flakiness?

Not that I'm opposed to the change. Although I would prefer if we could keep the indent at 4 spaces (for backward compatibility) when pretty printing. That doesn't seem to be too hard, I've created a fork of my own to play around with this. (But my GitLab build seems to have an issue getting docker base images at the moment)

chrnndz3:
@fhoeben 

> I never had any problems with the ordering of the results in my jUnit tests, or acceptance tests I wrote with this helpers (not locally, nor on CI servers). On what platform and with which JVM did you encounter the flakiness?
> 
> Not that I'm opposed to the change. Although I would prefer if we could keep the indent at 4 spaces (for backward compatibility) when pretty printing. That doesn't seem to be too hard, I've created a fork of my own to play around with this. (But my GitLab build seems to have an issue getting docker base images at the moment)

We use this tool called Nondex. We use it to detect and debugging wrong assumptions on under-determined Java APIs. When running these tests on different machines, the hardcoded values might come in different orders this will cause some tests to fail. As a result, after making the changes, it will then give the same results in all environments/ machines. Here's the research paper that explains it in details and how it works: https://www.cs.cornell.edu/~legunsen/pubs/GyoriETAL16NonDexToolDemo.pdf

I hope this helps clear up some clarification.

chrnndz3:
Hi @fhoeben, I'm following up on this PR. Please let me know if you need more information.

fhoeben:
Sorry I have not followed up sooner.

I have some performance optimisations, do you mind if I push those into this branch?

Furthermore I still feel it is a pity that you ended up making one of the unit test more robust, instead of fixing the underlying non-determinism for the sorting in json elements. But the changes already make other parts more deterministic so this PR is a step i the right direction.

PR #369, opened 2021-01-28
Webdriver updates

lennyg1:


pvreeuwijk:
ChromeDriver 88.0.4324.96 would be better, as Chrome 88 has been released

PR #373, opened 2021-03-30
Integration of shadow automation selenium 

Marc117:
This version of HSAC-Fitnesse-Fixtures allows the usage of shadow dom elements for test automation. It extends several methods of the BrowserTest class to use the [shadow automation selenium](https://github.com/sukgu/shadow-automation-selenium) by [sukgu](https://github.com/sukgu).

### DIFF
--- a/src/assembly/standalone.xml
+++ b/src/assembly/standalone.xml
@@ -53,7 +53,7 @@
                 <exclude>FitNesseRoot/RecentChanges/**</exclude>
                 <exclude>FitNesseRoot/RecentChanges.wiki</exclude>
                 <exclude>FitNesseRoot/ReRunLastFailures.wiki</exclude>
-                <exclude>FitNesseRoot/RerunLastFailures.wiki</exclude>
+                <exclude>FitNesseRoot/RerunLastFailures*.wiki</exclude>
                 <exclude>FitNesseRoot/TemplateLibrary/**</exclude>
                 <exclude>FitNesseRoot/properties</exclude>
                 <exclude>FitNesseRoot/updateList</exclude>


fhoeben:
Thanks for the PR!

The actual changes to BrowserTest seem very minimal, which is quite good.
Of course, I've not had the time to look in detail, but some initial observations:

- I don't believe I see any actual tests using the shadow dom features. You added a suite, outside the examples and acceptance tests, but I don't believe it's empty. Could you add some acceptance (and example) tests to show the feature in action?
- Does the shadow dom handling only work with technical selectors, or also with heuristics? Does it handle heuristic searches that don't find anything nicely? 
- Currently the shadow dom handling is always on (if elements are not found immediately), could we make that configurable (like angular or iframe handling) so that people can disable it if they know they don't need it (to keep tests quicker)
- Your code needs a bit of cleanup (no printing of stack traces, no `.*` imports, changes to existing files that probably are not needed in the project) ;-)

Marc117:
Hey, thank you for your feedback. I will try to correct the points you mentioned. Let me know if you any have further improvements and corrections.

Marc117:
Hey, 

after some time I think I did all the stuff you asked for:

- acceptance (and example) tests showing feature in action 
-> although to be fair there is not every method (which supports the shadow dom handling covered)
- configurable shadow dom handling
- cleanup of code

And to answer your question the plug in only supports technical selectors (CSS selector) to search for elements. I won't have the time for now to do more examples with all the other supported BrowserTest methods, so I hope this is enough. Thanks!

Completly unrelated to the inital PR I added a custom method that we needed at my work for our project. It is a simple method that lets you encode URLs. I know that there is already a url encode method somewhere in your fitnesse project, but we needed it in the BrowserTest class. You can keep it or delete it, however you like.

fhoeben:
@Marc117 Thanks for following up. I'll try to do a proper review this weekend.

Please remove the `urlEncode()` method. Since it is unrelated it should be a PR of its own. I don't believe there already is a method that does exactly this (can you explain how you use this? there is something similar in the implementation of HttpTest) but to me such a method would be a better fit for StringFixture, and the implementation should probably just use one of Java's URL encode methods instead of creating a URI and then getting the query part.

Marc117:
> Please remove the `urlEncode()` method. 

Yes I will do.

> can you explain how you use this? there is something similar in the implementation of HttpTest) but to me such a method would be a better fit for StringFixture, and the implementation should probably just use one of Java's URL encode methods instead of creating a URI and then getting the query part.

There is an example test page in the Fitness Root. But it works pretty simple. You just give over a string like this:
`|$encodedName= | url encode | Audi AG|`
The method returns the URL encoded version. For example "Audi AG" will get encoded to "Audi%20AG". The reason why I created the URI instead of using Java's URL encode method is that these use a different type of URL encode called x-www-form-urlencoded type and encode a space as "+" and not as "%20" as we would have needed it. But if you create an URI and get the rawQuery part its the right encoding type (percent-encoding).

PR #379, opened 2021-04-23
Post and get file

testsmith-io:
This is a new keyword that returns a binary file on a POST request.


fhoeben (changes requested):
You are sending the current values only as query string parameters, not as body. That doesn't seem right. Otherwise only small remarks.

I'mm create a PR to your branch so you can verify whether my cleanup still works for your situation

PR #380, opened 2021-05-04
Urlencode for String fixture class

Marc117:
As discussed, the url encode in its own PR. Not sure how to better handle the URISyntaxException, open for feedback. Thanks!


fhoeben:
Looks pretty good.

Can you add some proper whitespace before the `{`s?
And maybe it would be good to also offer a url decode?

PR #383, opened 2021-06-24
Shadow automation

Marc117:
Cleaned up my branch system in which everything was in master branch. To avoid confusion whats included I created a new branch for the shadow automation plugin and resulting with that a new PR. No new changes to original feature. Appreciate feedback.


src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -86,6 +96,14 @@ protected void beforeInvoke(Method method, Object[] arguments) {
         waitForAngularIfNeeded(method);
     }
 
+    public void analyzeLog(){
+        List<LogEntry> entries = driver().manage().logs().get(LogType.PERFORMANCE).getAll();
```
fhoeben, line 100: Why is this method here in this PR?
I actually like the ability to access the browser logs. But please put it in a separate PR. And the result should probably be sent to the test result page(as `ArrayList`?)  instead of standard out. And the ability to make assertions about the content of the logs would probably be good. 
(I've actually had the question whether we could make assertions on the absence of JavaScript errors on pages a couple of times...)

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -2732,6 +2756,65 @@ public void setNgBrowserTest(NgBrowserTest ngBrowserTest) {
...
+
+    protected T getElementWithShadow(String cssSelector){
+        try{
+            return (T) shadow.findElement(cssSelector);
+        } catch (Exception e){
+            return null;
+        }
+    }
+
+    private boolean ShadowUsageAllowed(T elementOrContainer) {
```
fhoeben, line 2775: Please make protected and use a method name in line with standard Java conventions (e.g. `isShadowUsageAllowed`). Looking at the body it is not immediately clear to me what the method does/returns.

fhoeben (changes requested):
I haven't quite gotten through the PR. But some initial remarks.

Normally BrowserTest requires a `css=` prefix for places based on CSS selector. You should also require that for these methods (i.e. only try shadow retrieval if prefix is present and remove the prefix before using the place).

Maybe this whole behaviour should be in a subclass (like there is for Angular with NgBrowserTest) instead of in the main class

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
Marc117: That should not be in the PR. My bad. Gonna remove it and maybe will be an separate PR if I finish it.

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
Marc117: This method checks if all conditions are given before using shadow. e.g. is shadow usage enabled and did fitnesse try to find an element several times the "normal" way before using the shadow plugin? If so and also if it still can't find anything after the 10th try it will return true.

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
```diff
@@ -2732,6 +2746,67 @@ public void setNgBrowserTest(NgBrowserTest ngBrowserTest) {
...
+        return shadowDomHandling;
+    }
+
+    public void setShadowDomHandlingTo(boolean shadowDomHandling){
+        this.shadowDomHandling = shadowDomHandling;
+    }
+
+    protected T getElementWithShadow(String place){
+        try{
+            if(place.startsWith("css=")) return (T) shadow.findElement(place.split("css=")[1]);
```
Marc117, line 2759: It now requires the "ccs=" prefix to work with shadow. It is probably not the best way to do it and doesn't use the SeleniumHelper methods but I couldn't figure it out how to make that work. Open for better ideas.

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
Marc117: Changed it to "isShadowUsageAllowed".

Marc117:
I addressed everything you mentioned. The "css=" prefix is now required to find an element with shadow. Although it is for sure not the best way to do it and as stated in the comment regarding that line open for better ideas. One thing I was not sure about if I should try do it or not:
> Maybe this whole behaviour should be in a subclass (like there is for Angular with NgBrowserTest) instead of in the main class

In case it is really needed to do it that way I can look if I find time to realize it but wanted to make sure first.

src/main/java/nl/hsac/fitnesse/fixture/slim/web/BrowserTest.java
Marc117: Removed it.

PR #398, opened 2022-01-31
Improve execute program

maartenjanvangool:
Two items I ran into while using this fixture. 
1. A method to check if the output contains something. 
2. The actual issue logged when starting a program does not work, instead of only a generic message.


fhoeben:
@maartenjanvangool I never added a method to check for standard out contains as I believe the same effect (and much more) can be achieved by using a regular expression as the expected value in a check of the output. Do you see any additional benefits? If so: can you add a similar method for standard error output, and add some tests to verify the behaviour?

I believe the actual message was already present in the generated stack trace, but one had to scroll down to the 'cause' to see it. Having it more at the top makes sense.

PR #355, opened 2020-10-26
Selenium 4 Alpha

tcnh:
This PR contains a whole bunch of things:
 * Selenium 4 Alpha 6, allowing for:
    * ChromiumDrivers/Options usage. Which is implemented in LocalDriverFactory. Some examples of chromium (Chrome/Modern Edge configuration) are in the wiki. Should make configuration more straightforward.
    * Relative Locators - Click/Enter as methods as examples are in BrowserTest. (i.e. `| click | input | left of | bad coded checkbox label |`
    * DevTools API for chromium (examples in the wiki files). I.E:
      * Request Header injection
      * Basic/Digest authentication
      * Cookie injection
      * Geo Location Mocking
      * (XHR) Response overrides/mocking
      * Request Logging
      * Performance Metrics
      * Full page screen shots

This is not production-ready (the alpha status of Selenium 4 with all of its quirks alone proves that), but I'd like to see some review/testing it in a bit of a broader perspective than just my own. I believe the possibilities (especially leveraging devTools) can prove valuable.

PR #400, opened 2022-02-06
Selenium 4

tcnh:
New Selenium 4 PR. (continues on base of https://github.com/fhoeben/hsac-fitnesse-fixtures/pull/355)

Selenium 4.1.2 version
Relative locators
Shadow DOM access by using existing searchcontexts in Chromium 96+
DevTools Browsertest as a sample
ChromiumOptions for Chrome- and Edgedriver, with examples wiki page

Includes acceptance tests for shadowdom and relative locators


pom.xml
```diff
@@ -43,8 +43,8 @@
 		<hsac-fitnesse-plugin.version>1.31.0</hsac-fitnesse-plugin.version>
 		<fitnesse.version>20211030</fitnesse.version>
 		<praegus-toolchain-fitnesse-plugin.version>2.0.14</praegus-toolchain-fitnesse-plugin.version>
-		<selenium.version>3.141.59</selenium.version>
-		<slf4j.version>1.7.32</slf4j.version>
+		<selenium.version>4.1.2</selenium.version>
+		<slf4j.version>1.7.25</slf4j.version>
```
fhoeben, line 47: Any particular reason to downgrade slf4j?

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/Selenium4Examples.wiki
```diff
@@ -0,0 +1,3 @@
+!1 Test suite X
```
fhoeben, line 1: I suppose this file is no longer used?

fhoeben (commented):
Haven't gone in full details yet, but looks pretty good.

Some global thoughts:
- should we add some documentation (references) to the top-level examples.BrowserTest.wiki for the new commands/features?
- the relative locators add a lot of new methods to the (already too big) BrowserTest class. And the require technical selectors (which you very nicely explicitly documented). Can we do something about that? I'm thinking whether it would be bad to have them in a separate subclass (I like that they are available, but think I would prefer tests that do not (need to) use them). Do you have any thoughts?

pom.xml
tcnh: was meant to be .35 but I think i typoed. Will update

tcnh:
> Haven't gone in full details yet, but looks pretty good.
> 
> Some global thoughts:
> 
> * should we add some documentation (references) to the top-level examples.BrowserTest.wiki for the new commands/features?

I added them to the commands table (value of <> above/below/leftof/right of/near was missing, is added now)
I feel giving them much more acknowledgement would be to much credit for methods that should relatively not be used very often?

> * the relative locators add a lot of new methods to the (already too big) BrowserTest class. And the require technical selectors (which you very nicely explicitly documented). Can we do something about that? I'm thinking whether it would be bad to have them in a separate subclass (I like that they are available, but think I would prefer tests that do not (need to) use them). Do you have any thoughts?

I agree (see above) that these are a bit of niche. On the other hand this is the most prominent addition to Selenium 4 for most end-users and I also believe there are a whole lot of methods in BrowserTest that are only used in specific use cases, so these are  not that different. Splitting browsertest into multiple layers (of complexity/technology orientation/etc) might be a good idea, but as I mentioned, I believe we could move out a whole bunch of other methods and settings, too.

PR #422, opened 2022-08-08
update chromium based webdrivers to v104

timovd:
ChromeDriver 104.0.5112.79
EdgeDriver 104.0.1293.47


fhoeben:
I believe this was covered by #421

PR #462, opened 2023-01-16
ChromeDriver/EdgeDriver update

lennyg1:


fhoeben:
Looks like GitHub action runner is not yet updated to Chrome 109. It probably will be soon.

Can you also update `changelog.md`?

fhoeben:
I was able to install chrome 109 manually in another branch. If they don't update the runner soon I'll add that and merge this PR anyway

PR #470, opened 2023-02-10
Update chromium based webdrivers to v110

timovd:
- ChromeDriver 110.0.5481.77
- Edgedriver 110.0.1587.41


timovd:
This version of ChromeDriver only supports Chrome version 110
Current browser version is 109.0.5414.119 with binary path /usr/bin/google-chrome 

Wil try again later...

timovd:
@fhoeben could you rerun the tests? Seems like GitHub was still on an older version.

fhoeben:
@timovd build is green now

PR #490, opened 2023-04-24
Selenium 4.9.0 and JsonPath 2.8.0

lennyg1:


fhoeben:
Thx @lennyg1 , could you take a closer look at the diff in `pom.xml`? I believe you changed a bit more than required. Mostly whitespace I believe, but you also introduced `json-smart`?

lennyg1:
I had to add the dependency, because otherwise I'd get a compilation error. Json-smart should've been updated in [JsonPath 2.8.0](https://github.com/json-path/JsonPath/blob/master/build.gradle), however somehow it doesn't want to work.

Yes, I also formatted the pom.xml after adding the dependency. I can revert that if you prefer.

lennyg1:
@fhoeben apparently json-path changed the dependency scope to runtime. However adding the dependency to dependencymanagement with a different scope fixes this. The PR has been updated.

PR #516, opened 2023-06-12
Webdriver and changelog update

lennyg1:


fhoeben:
@lennyg1 Thanks. Guess we again have to wait for GitHub to update its runners to have the check become green....

PR #521, opened 2023-07-24
New chromedriver endpoint and new webdrivers

lennyg1:


lennyg1:
Also updated Selenium to 4.11.0.

lennyg1:
Doesn't work on Windows due to it not detecting new chromedriver.

PR #544, opened 2023-10-09
EdgeDriver Mac

lennyg1:


changelog.md
```diff
@@ -21,12 +21,15 @@ An example of Maven based project using this baseline can be found at: https://g
...
+- EdgeDriver for Mac M1, Mac64, Linux
+- EdgeDriver 117.0.2045.55
+- ChromeDriver 
 
 New in 5.2.45
 - Selenium 4.13.0
 - The project no longer includes the Selenium webdriver for Internet Explorer
 - EdgeDriver 117.0.2045.40
-- ChromeDriver 117.0.5938.92
+- ChromeDriver 117.0.5938.149
```
fhoeben, line 32:
```suggestion
- ChromeDriver 117.0.5938.92
```

changelog.md
```diff
@@ -21,12 +21,15 @@ An example of Maven based project using this baseline can be found at: https://g
 When upgrading from an older version, ensure all Selenium webdrivers are stopped and delete the 'wiki/plugins' and 'wiki/webdrivers' folders to remove old versions of the packaged plugins/webdrivers. When using the sample project: follow the instructions at https://github.com/fhoeben/sample-fitnesse-project#upgrading.
 
 New in ${VERSION}
+- EdgeDriver for Mac M1, Mac64, Linux
+- EdgeDriver 117.0.2045.55
+- ChromeDriver
```
fhoeben, line 26:
```suggestion
- ChromeDriver 117.0.5938.149
```

fhoeben (changes requested):
Based on the exclude https://github.com/fhoeben/hsac-fitnesse-fixtures/blob/cc58b071b6807d0b924991bb9f2762a9a27bb385/src/assembly/standalone.xml#L39C16-L39C56 I don't believe the Linux driver will be added to the standalone.zip distribution and as such I don't believe it should be mentioned here.

Do you see (a lot of) demand for testing with Edge on a Mac instead of with Chrome? Adding these drivers makes the standalone.zip larger yet. In the past I saw little push to test on anything other than Safari and Chrome for Macs from customer organizations and given the Edge and Chrome share their engine I would expect too much added value from testing with Edge on a Mac.
If I were to test specifically with Edge I would do that for Windows machines, since that's where I expect actual usage to be.

Are there big advantages testing using Edge on Mac? Or is this something you experience customer demand for?

lennyg1:
Chrome 119 has been released, that means that the webdrivers within HSAC are giving errors now if you use the latest Chrome. Is it possible to accept this PR?

fhoeben:
Can you remove the addition of the Mac Edge drivers, so we just have the version update? I'm not convinced yet. (And I think changing the version and adding other platforms is better kept in separate PRs anyway)

lennyg1:
> Can you remove the addition of the Mac Edge drivers, so we just have the version update? I'm not convinced yet. (And I think changing the version and adding other platforms is better kept in separate PRs anyway)

I've had the question a few times now due to Edge increasing in popularity. Even on Linux. Still an edge case, but not impossible anymore. I'll make a new PR.

fhoeben:
@lennyg1  I'm working on providing a separate artifact which includes the web drivers and one without. In the one with I'm planning to also add the Edge drivers for Mac

PR #552, opened 2023-11-02
Webdriver updates

lennyg1:


fhoeben:
Sorry, you're too new for GitHub.
Should we make a version for Chrome 118 (and then a new PR for 119), or just wait until GitHub runners are updated

PR #556, opened 2023-11-07
Selenium webdriver and plugin update

lennyg1:


lennyg1:
Might be possible to manually update the browser by adding the lines below to other-java11.yml on line 34.

      - name: Update browser
        run: |
          wget -q -O /tmp/chrome.deb https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb \
          && sudo apt install -y /tmp/chrome.deb \
          && rm /tmp/chrome.deb

fhoeben:
@lennyg1 annoyingly GitHub still hasn't updated their runners to include Chrome 119.

Your fix could work, or we could see whether we can use https://github.com/browser-actions/setup-chrome to ensure we install the right chrome. I would prefer to install a controlled version instead of 'stable current', and I believe that action supports that. I did see in [one of the issues](https://github.com/browser-actions/setup-chrome/issues/329) that it installs chrome in a non-default location (`/opt/google/chrome`), so we would have to ensure we add the location to the chrome options.

lennyg1:
To me the 'right' Chrome is always stable current. Due to people paying more attention to security, browsers simply get updated quicker and people figured out that end-users are also almost always on the latest browser versions anyway. Those with different demands are better off with a browser in a specific container anyway. I personally almost always use the containers with browser from Selenium itself.

fhoeben:
For the 'main' branch I agree with you, but for stable builds (e.g. we want to patch an older release) it is better to have explicit control

fhoeben:
Before I can merge this you also have to update the other 'java-11' workflow, so the updated browser is also available during the release workflow

PR #593, opened 2024-01-03
new examples based on v4 of practicesoftwaretesting.com

roydekleijn:


wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/ScriptTest.wiki
```diff
@@ -1,42 +1,38 @@
...
-|wait                                   |1                 |seconds                                            |
-|show                                   |take screenshot   |cover                                              |
+|script          |browser test                                                            |
+|open            |https://v4.practicesoftwaretesting.com                                  |
+|enter           |Pliers                              |as|Search                          |
+|press           |enter                                                                   |
+|wait for visible|css=[data-test='search_completed']                                      |
+|click           |Combination Pliers                                                      |
+|click           |Add to cart                                                             |
+|click           |css=[data-test='nav-cart']                                              |
```
fhoeben, line 18: This, of course, works. But maybe for the 'accessibility' using screen readers of your site you should add an aria-label to the cart icon.
The element could then be found using this label...

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/TableTemplateTest.wiki
```diff
@@ -1,34 +1,33 @@
...
+|click                                  |Combination Pliers                    |
 
-Then we check some entries.
+Then we check the related products.
 
-|table template|album details                                |
-|scroll to     |@{album}                                     |
-|$year=        |value of|Year |in row where|Album|is|@{album}|
-|$label=       |value of|Label|in row where|Album|is|@{album}|
+|table template|related products     |
```
fhoeben, line 21: As stated above, with only a single parameter this example doesn't really show the point I believe.
Having at least 3 would be my preference (for instance product, price and brand?)

fhoeben (changes requested):
Thanks Roy a good start, but I think we can make it even nicer with some additional changes.

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/ScriptTest.wiki
testsmith-io: yep, makes sense...

roydekleijn (commented):
addressed comments

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/CustomFixtureTest.wiki
```diff
@@ -16,36 +16,35 @@ There are some things a Java method can do that a scenario cannot: for instance
...
-|click                                  |Van Morrison                              |
-|wait for visible                       |Album Highlights                          |
+|script          |browser test                                                      |
+|open            |https://v4.practicesoftwaretesting.com                            |
+|click |Sign in                                              |
+|enter |customer@practicesoftwaretesting.com|as|Your E-mail  |
+|enter |welcome01                           |as|Your Password|
+|click |Login                                                |
+|click           |Profile                                                           |
+|wait for visible|Update Profile                                                    |
```
fhoeben, line 26: Please fix alignment of `|` characters

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/ScenarioTest.wiki
```diff
@@ -1,37 +1,37 @@
...
+Then we check the related products.
 
-|scenario |album details _|album, year?, label?                |
-|scroll to|@{album}                                            |
-|$year=   |value of       |Year |in row where|Album|is|@{album}|
-|$label=  |value of       |Label|in row where|Album|is|@{album}|
+|scenario  |related products _|product, price?, brand?|
+|click     |@{product}                                   |
+|$price=   |value of          |unit-price                |
+|$brand=|value of          |brand                  |
```
fhoeben, line 26: `|` alignment

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/TableTemplateTest.wiki
```diff
@@ -1,34 +1,34 @@
...
+|click     |@{product}                                   |
+|check|value of          |unit-price                |@{price}|
```
fhoeben, line 23: Please use the syntax where the values are assigned to a Slim variable which is then compared to the supplied value in the scenario table (this gives much nicer output if the values do not match).

```suggestion
|$price=|value of          |unit-price    |
```

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/TableTemplateTest.wiki
```diff
@@ -1,34 +1,34 @@
...
+|check|value of          |brand                  |@{brand}|
```
fhoeben, line 24:
```suggestion
|$brand=|value of          |brand                  |
```

wiki/FitNesseRoot/HsacExamples/SlimTests/BrowserTests/TableTemplateTest.wiki
```diff
@@ -1,34 +1,34 @@
...
 
 
-|album details                          |
-|album       |year?|label?              |
-|Astral Weeks|1968 |Warner Bros.        |
-|Moondance   |1970 |Rhino / Warner Bros.|
+|related products |
+|product          |price|brand|
```
fhoeben, line 28:
```suggestion
|product          |price?|brand?|
```

fhoeben (changes requested):
Almost there.
Just ensure to (auto) format all example pages so the `|` characters are aligned, and use the nicer syntax in the table template example

fhoeben (changes requested):
You didn't address my comments on table template example...

fhoeben:
I just ran the tests in the full suite, and I noticed that the pages that login fail if a previous test in the suite already logged in.
Can we easily prevent that (for instance by 'delete all cookies' at the start of the tests, or by having a tear down that signs out)?
Screenshot 2024-01-05 at 14 45 37


Furthermore the custom fixture test failed again and the scenario and table template tests seem to have the wrong prices?

Screenshot 2024-01-05 at 14 53 49

fhoeben:
At the moment it appears I can make the custom fixture test pass by changing the first browser test based table to:
```
|script|browser test                                         |
|open  |https://v4.practicesoftwaretesting.com               |
|click |Sign in                                              |
|enter |customer@practicesoftwaretesting.com|as|Your E-mail  |
|enter |welcome01                           |as|Your Password|
|click |Login                                                |
|click |Profile                                              |

!2 Using !-BrowserTest-!

|script|browser test        |
|click |City                |
|check |value of|City|Vienna|
```
Then I don't even need the 'wait for visible' in the table above it...

fhoeben:
Actually it turns out the whole custom fixture is doing the wrong thing for this page. It should be detecting that the javascript has updated the 'shadow dom' for the inputs, because once that is done  the table below just works (no need for xpath or custom java logic at all)
```
|script|browser test            |
|check |value of|City   |Vienna |
|check |value of|Country|Austria|
```

Which JavaScript framework are you using? When I throttle the network bandwidth in chrome dev tools I also clearly see a minimal page first, than the global template of the page is loaded (all inputs are empty and the 'Jane Doe' menu on the top left corner is not yet present, then that menu is added and then the content is added to the inputs. Is there some Javascript command we can execute to check whether all is done, or some element that only appears after the whole rendering is complete?

roydekleijn:
I use Angular.

image

fhoeben:
I guess you are using Angular. I can make the test more stable by using Ng2BrowserTest, which includes Javascript to ensure the page is rendered fully before checking values.

fhoeben:
@roydekleijn I tried to improve the test (and Java code) a bit, what do you think?

roydekleijn (commented):
Looks good to me!

PR #597, opened 2024-01-11
Shorthand method to start a selenium-managed driver

tcnh:
Since selenium 4.6.0, Selenium contains its own [driver manager](https://www.selenium.dev/blog/2022/introducing-selenium-manager/) that can automatically download the right webdriver version for a given browser. This manager can currently be used with selenium driver setup using the `|start driver|<driver class FQN>|` methods (or override from maven using the driverclass.

This, however, is not immediately clear to all users and using the right driver classname is apparently not easy for everyone..
This PR introduces new methods (startManagedDriverFor & startManagedDriverForWithProfile) that derive the driver class name from the readable driver name (chrome, edge, firefox, safari, internet_explorer) and make it explicit that a **selenium managed** driver is desired.

Now, when users would set `<webdriver.download.skip>true</webdriver.download.skip>` in their project's pom.xml and use these methods, they will no longer run into issues with web driver versions getting out of sync when their company laptop auto-updates (or never updates) the browser.

The selenium driver manager works on windows, linux and mac.
Also added this driver usage in the Example SuiteSetUp for BrowserTest to both illustrate the option and prevent users from not being able to run the examples due to browser vs driver incompatibilities.

### DIFF
--- a/src/main/java/nl/hsac/fitnesse/fixture/slim/web/SeleniumDriverSetup.java
+++ b/src/main/java/nl/hsac/fitnesse/fixture/slim/web/SeleniumDriverSetup.java
@@ -4,6 +4,7 @@ import nl.hsac.fitnesse.fixture.Environment;
 import nl.hsac.fitnesse.fixture.slim.SlimFixture;
 import nl.hsac.fitnesse.fixture.util.selenium.SauceLabsHelper;
 import nl.hsac.fitnesse.fixture.util.selenium.SeleniumHelper;
+import nl.hsac.fitnesse.fixture.util.selenium.driverfactory.DriverClassNames;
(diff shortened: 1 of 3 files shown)


fhoeben:
I really appreciate this new feature, I was thinking we should support the managed drivers since I learned of that Selenium feature.

One thing I miss in the PR is an update of the `changelog.md`. 
Furthermore I'm not sure whether we should make the new option the default for the examples. 

In general I expect people to use the sample project setup for their own tests. There the drivers are never downloaded from their providers (which is what is configured here using `webdriver.download.skip`) but they are taken from our 'standalone.zip'. So if we make the managed drivers the default we should probably make separate artifacts with and without the web drivers, so the basic download does not have to include the drivers.

PR #632, opened 2024-04-09
dependency updates and webdrivers

lennyg1:
Webdrivers updated, new urls for ChromeDriver (again). Selenium upgraded and some Maven plugins.


lennyg1:
Removed richfacestest, it's been EOL since 2016 (https://richfaces.jboss.org/) and the tests now fail because the Tridos website in the example/acceptance test doesn't seem to use it anymore and I can't find a replacement.

fhoeben:
@lennyg1 maybe the test site is/was broken (again) or the test is broken, but I just visited the test site and it still seems to use RichFaces (at least that what I think the `seam` in the URL indicates).
Having said that the test is probably not very relevant anymore as hardly anyone is still using this framework.

fhoeben:
Used these changes as basis for separate commit to web drivers 124 and selenium 4.20

Issue #675, opened 2024-10-17
Use slf4j to log errors and information not System.out / System.error


e-roeling:
We use elastic search to collect the logging of all environments. For environments on which FitNesse is running we run into issues with stack traces, this is caused by the fact that they are just printed to System.out.

I see that some classes already make use of slf4j. It would help us if this can be done for all classes (Except for JUnitTeamcityReporter as the logging in this class if from a different nature.)

I will add a pull request for this.

PR #676, opened 2024-10-17
[675] Log info and errors using slf4j

e-roeling:


e-roeling:
See https://github.com/fhoeben/hsac-fitnesse-fixtures/issues/675#issue-2593789875

src/main/java/nl/hsac/fitnesse/junit/ProgressLoggerListener.java
fhoeben, line 1: I don't think this should be updated as it will be used during jUnit run. I doubt slf4j is desirable there

src/main/java/nl/hsac/fitnesse/junit/JUnitConsoleRunner.java
fhoeben, line 1: I don't think this should be updated as it will be used in `main` method launching the process from the command line. That should just use std error

src/main/java/nl/hsac/fitnesse/junit/HsacFitNesseRunner.java
fhoeben, line 1: I don't think this should be updated as it will be used during jUnit run. I doubt slf4j is desirable there

fhoeben (changes requested):

src/main/java/nl/hsac/fitnesse/junit/ProgressLoggerListener.java
e-roeling: I don't fully agree, but can live with it as its quite simple to use our own implementation for these.

fhoeben (approved):

PR #777, opened 2026-04-13
Selenium 4.43

tcnh:


fhoeben:
@tcnh two test failure I also noticed earlier trying to update to a more recent Selenium.
Do you have the time to investigate those?

tcnh:
It seems either the definistion of visible has improved, or chromium is more picky about non-closed html tags. I think it's the first.. When we ensure some actual space is rendered, both tests pass as expected

fhoeben:
Thanks a lot, I was hoping it would be something small like that but didn't have time to look at it yet.

One last ask, can you also add this upgrade to the changelog

